PR #105, opened 2022-02-24
Persist /trackInAppClick When Navigating Away

martinmckenna:
## JIRA Ticket(s) if any

* [MOB-4071](https://iterable.atlassian.net/browse/MOB-4071)

## Description

Previously, when a user opened an in-app message and clicked on a link that was going to navigate them to a new URL in the same tab, the `/trackInAppClick` request would be called, but never finish because the page was being unmounted.

Now, we persist the request using the `fetch` API's `keepalive` flag. Essentially what it does is makes sure the request finishes even though the user navigated to a new page.

🚨  PLEASE NOTE: In the Devtools on your browser, the request will be labeled as "unknown" because the browser doesn't actual know if the request succeeded or failed, and neither do we. We just simply hope for the best 🚨 

Screen Shot 2022-02-24 at 1 35 21 PM

## Test Steps

1. Send yourself a (non-proof) in-app message with a link that will navigate you in the same tab
   * e.g. `<a href="/about">hello</a>` or `<a href="localhost:8080/something">hello</a>`
2. Run the example app with `yarn install:all && yarn start:all`
3. Login and paint the in-app message to the screen
4. Click your same-tab link and then head over to https://app.iterable.com/users/lookup and find your user
5. Confirm that a click event was tracked even though you navigated away

### DIFF
--- a/src/authorization/utils.ts
+++ b/src/authorization/utils.ts
@@ -1,4 +1,5 @@
 import { Buffer } from 'buffer';
+import axios, { AxiosRequestConfig } from 'axios';
 
 export const getEpochExpiryTimeInMS = (jwt: string) => {
   /** @thanks https://stackoverflow.com/a/38552302/7455960  */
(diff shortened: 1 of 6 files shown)


src/authorization/utils.ts
```diff
@@ -44,3 +45,44 @@ export const getEpochDifferenceInMS = (
   }
   return parsedFuture - parsedNow;
 };
+
+export const cancelAxiosRequestAndMakeFetch = (
+  config: AxiosRequestConfig,
+  { email, userID }: { email?: string; userID?: string },
```
pauljung14, line 51: @martinmckenna just curious as to why this param is an object like so? It's not apparent to me the benefit of putting them together like this unless I'm missing something?

pauljung14 (approved):
Lgtm, just had one question about a param

src/authorization/utils.ts
martinmckenna: so when the customer's code identifies the user (either by email or userID), the identification method they chose will determine what's defined here. So it the customer wrote code like this:


```ts
import { initialize } from '@iterable/web-sdk'

const { setEmail } = initialize(...whateverArgs);

setEmail('email@email.com')
```

then here in this function we'll have an email, but not a user ID. They're in an object like this just so we can be explicit when we're consuming them.

I suppose we could just have one argument called `emailOrUserID` but I like the idea of knowing what we're working with and having an explicit name.

The alternative being:

```ts
import { initialize } from '@iterable/web-sdk'

const { setUserID } = initialize(...whateverArgs);

setUserID('1fdws343')
```

martinmckenna: but to your point, we totally could take them like this:

```ts
export const cancelAxiosRequestAndMakeFetch = (config: AxiosRequestConfig, email?: string, userID?: string) => {}
```

I just prefer if we had one master "identification information" object 🤷‍♂️

PR #123, opened 2022-04-21
Add More Display Options to getMessages 

martinmckenna:
## JIRA Ticket(s) if any

* [MOB-4296](https://iterable.atlassian.net/browse/MOB-4296)

## Description

Updates `getMessages` API to allow for deferring the display of in-app messages to any time the user wants. API looks like this:

```ts
const { request, triggerDisplayMessages } = getMessages(...options, { display: 'deferred' })

request().then(response => triggerDisplayMessages(response.data.inAppMessages))
```

## Test Steps

1. Run a local JWT generation backend by [cloning this repo](https://github.com/Iterable/jwt-generator) and running `docker-compose up -d`
2. Run example app in this repo with `yarn install:all && yarn start:all`
3. Open up `localhost:8080` in your browser and login
    * change you email in the codebase by replacing `iterable.tester@gmail.com` with your email
4. In `src/example/index.ts` change `getMessages` call to these variants:

```ts
/* should not display any messages */
getMesssages(...options).then(response => response)
```

```ts
const { request } = getMesssages(...options, true)

/* ...etc */

/* should display messages immediately */
request()
```

```ts
const { request } = getMesssages(...options, { display: 'immediate' })

/* ...etc */

/* should display messages immediately */
request()
```

```ts
const { request, triggerDisplayMessages } = getMesssages(...options, { display: 'deferred' })

/* ...etc */

request().then(response => {
   /* should only display messages when _triggerDisplayMessages_ is called */
  triggerDisplayMessages(response.data.inAppMessages)
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -598,7 +598,7 @@ import { baseAxiosRequest } from '@iterable/web-sdk/dist/request';
 
 ## I want to automatically show my in-app messages with a delay between each
 
-This SDK allows that. Simply call the `getMessages` method but pass `{ display: 'immeidate' }` as the second parameter. This will expose some methods used to make the request to show the messages and pause and resume the queue.
+This SDK allows that. Simply call the `getMessages` method but pass `{ display: 'immediate' }` as the second parameter. This will expose some methods used to make the request to show the messages and pause and resume the queue.
 
 Normally to request a list of in-app messages, you'd make a request like this:
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Below are the methods this SDK exposes. See [Iterable's API Docs](https://api.it
 
 | Method Name           	| Description                                                                                                               	|
 |-----------------------	|---------------------------------------------------------------------------------------------------------------------------	|
-| [`getInAppMessages`](#getInAppMessages)    	| Either return in-app messages as a Promise or automatically paint them to the DOM if the second argument is passed `true` 	|
+| [`getInAppMessages`](#getInAppMessages)    	| Either return in-app messages as a Promise or automatically paint them to the DOM if the second argument is passed `DisplayOptions` 	|
 | [`initialize`](#initialize)        	| Method for identifying users and setting a JWT                                                                            	|
 | [`track`](#track)               	| Track custom events                                                                                                       	|
 | [`trackInAppClick`](#trackInAppClick)     	| Track when a user clicks on a button or link within a message                                                             	|
@@ -135,7 +135,7 @@ const {
       topOffset: '20px'
     }
   },
-  true
+  { display: 'immediate' }
 );
 
 request()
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -60,7 +60,7 @@ Below are the methods this SDK exposes. See [Iterable's API Docs](https://api.it
 | [`updateCart`](#updateCart)          	| Update _shoppingCartItems_ field on user profile                                                                          	|
 | [`updateSubscriptions`](#updateSubscriptions) 	| Updates user's subscriptions                                                                                              	|
 | [`updateUser`](#updateUser)          	| Change data on a user's profile or create a user if none exists                                                           	|
-| [`updateUserEmail`](#updateUserEmail)     	| Change a user's email address                                                                                             	|
+| [`updateUserEmail`](#updateUserEmail)     	| Change a user's email and reauthenticate user with the new email address (in other words, we will call `setEmail` for you)                                                                                             	|
```
martinmckenna, line 64: view the README in a prettier way here:

https://github.com/Iterable/iterable-web-sdk/blob/3dee5eddea3fd605030f243956776413dfb593fe/README.md

README.md
```diff
@@ -69,7 +69,7 @@ Below are the methods this SDK exposes. See [Iterable's API Docs](https://api.it
 API [(see required API payload here)](https://api.iterable.com/api/docs#In-app_getMessages):
 
 ```ts
-getInAppMessages: (payload: InAppMessagesRequestParams, showMessagesAutomatically?: boolean) => Promise<TrackConsumeData> | PaintInAppMessageData
+getInAppMessages: (payload: InAppMessagesRequestParams, showMessagesAutomatically?: boolean | { display: 'deferred' | 'immediate' }) => Promise<TrackConsumeData> | PaintInAppMessageData
```
pauljung14, line 72: How do `true` and `false` behaviors differ from `{display: 'deferred'}` and `{display; 'immediate'}`? I.e., are there 4 different behaviors? Do you think it'd be worth providing a sentence in the readme to explain how they differ?

pauljung14 (approved):
lgtm! just had 1 question

README.md
martinmckenna: yes I can explain! essentially, these two do the same thing:

```
getMessages(options, true);
getMessages(options, { display: 'immediate' })
```

The only reason I'm keeping the boolean value is so that we remain backwards-compatible and this next release doesn't break any implementations that the customers have already written. In other words, I don't want customers to start getting errors if they go from `v0.1.3` -> `v0.2.0` (which is what the next release is going to be).

My plan is to follow the semantic versioning strategy and only bump the major version when breaking changes are introduced. So `true | false` values will be deprecated once there is another major version bump, but until then, we'll keep it around.

PR #142, opened 2022-08-16
[MOB-4576] Secure unprotected raw html path with iframe

ts-nguyen:
## JIRA Ticket(s) if any

(https://iterable.atlassian.net/browse/MOB-4576)

## Description

- Updates `getInAppMessages` path when `showInAppMessagesAutomatically` is not passed in and wraps the inapp message HTML in the API response with an iframe to be securely rendered to prevent JS execution

## Test Steps

Run the example app and pull messages without showing messages automatically and see that the HTML returned is instead an iframe with the HTML content inside and a sandbox attribute

### DIFF
--- a/src/inapp/utils.ts
+++ b/src/inapp/utils.ts
@@ -239,7 +239,7 @@ const mediaQueryXl = global?.matchMedia?.('(min-width: 1301px)');
 const generateSecuredIFrame = () => {
   const iframe = document.createElement('iframe');
   iframe.setAttribute('id', 'iterable-iframe');
-  iframe.setAttribute('sandbox', 'allow-same-origin');
+  iframe.setAttribute('sandbox', 'allow-same-origin allow-popups');
   /*
     _display: none_ would remove the ability to set event handlers on elements
     so instead we choose to hide it visibly with CSS but not actually remove
--- a/src/inapp/utils.ts
+++ b/src/inapp/utils.ts
@@ -239,7 +239,12 @@ const mediaQueryXl = global?.matchMedia?.('(min-width: 1301px)');
 const generateSecuredIFrame = () => {
   const iframe = document.createElement('iframe');
   iframe.setAttribute('id', 'iterable-iframe');
-  iframe.setAttribute('sandbox', 'allow-same-origin allow-popups');
+  // allow-popups and allow-top-navigation is to enable links for Safari since the iframe will block
+  // event handlers on elements in it preventing our custom link handling
+  iframe.setAttribute(
+    'sandbox',
+    'allow-same-origin allow-popups allow-top-navigation'
+  );
   /*
     _display: none_ would remove the ability to set event handlers on elements
     so instead we choose to hide it visibly with CSS but not actually remove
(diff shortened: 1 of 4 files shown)


src/inapp/utils.ts
```diff
@@ -232,6 +232,48 @@ const mediaQueryLg = global?.matchMedia?.(
...
+};
+
+/**
+ *
+ * @param html
+ * @returns { string } html string of an inapp message to wrap in an iframe for render
+ */
+export const wrapWithIFrame = (html: string): string => {
+  const iframe = generateSecuredIFrame();
+  return iframe.outerHTML.replace(
```
elarkin, line 269: You should be able to write to the iframe's document even before it's attached to the DOM, is that not reliable in all supported browsers?

ts-nguyen: When I was trying to write to the document it didn't exist so it wouldn't write, do you know which attribute I should be writing to? I tried both contentWindow.document and contentDocument but they were null until it was attached to the DOM

src/inapp/utils.ts
```diff
@@ -232,6 +232,48 @@ const mediaQueryLg = global?.matchMedia?.(
...
 const mediaQueryXl = global?.matchMedia?.('(min-width: 1301px)');
 
+/**
+ *
+ * @returns { HTMLIFrameElement } iframe with sandbox and hidden styling applied for of an inapp message to render with
+ */
+const generateSecuredIFrame = () => {
+  const iframe = document.createElement('iframe');
+  iframe.setAttribute('id', 'iterable-iframe');
+  iframe.setAttribute('sandbox', 'allow-same-origin allow-popups');
```
pauljung14, line 242: I think it'd be good to provide a short comment to justify why we're adding `allow-popups`, perhaps something like `allow-popups: Allows links with target="_blank" inside iframe to successfully open in a new tab.`

pauljung14 (approved):
See 1 comment

src/inapp/utils.ts
elarkin: i've been struggling with this technique. Stringifying code is often a source of vulnerability, but in this case I'm not seeing anything obviously.

That said, I also don't know that it's necessary to write the onload event as a string. Is it possible instead to use the form:

```
iframe.onload = function() { iframe.contentWindow.document.write(html); }
```

Is that something the team tried already? I am a little unsure regarding whether you get the closures given the iframe boundary, but it should be a little cleaner if it works. We'd want to see how IE/Edge handle it even if it works in Chrome.

ts-nguyen: So this function is just a transform function to take the raw HTML string and place it inside an iframe through the onload event. It hasn't reached the customer's app code yet so they still expect it as a string from the API response to render into their app. Since it's still technically in the API layer I don't think we want to mess with the response type and it should match our actual API response type which is just a string for the HTML which is why I've done it this way.

src/inapp/utils.ts
```diff
@@ -232,6 +232,53 @@ const mediaQueryLg = global?.matchMedia?.(
...
+/**
+ *
+ * @param html
+ * @returns { HTMLIFrameElement } iframe with html string of an inapp message wrapped in onload
+ */
+export const wrapWithIFrame = (html: string): HTMLIFrameElement => {
+  const iframe = generateSecuredIFrame();
+  iframe.onload = () => {
+    iframe.contentWindow?.document.write(
+      html.replaceAll(/\r?\n|\r/g, '').replaceAll('"', '\\"')
```
elarkin, line 276: You probably don't need to do this at all anymore. Calling write on (html) should work just fine. I suspect that the " to \" replacement will create some visual artifacts.

elarkin (approved):
other than functional issue noted

src/inapp/utils.ts
ts-nguyen: Yep my bad missed commiting that change

PR #154, opened 2022-09-07
[MOB-4800] Resolve dependency security vulnerability with terser

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4800](https://iterable.atlassian.net/browse/MOB-4800)

## Description

Need to upgrade terser to latest version to resolve security vulnerability. Terser is used up the chain by `html-webpack-plugin#html-minifier-terser` so this upgrade bumps versions for these.

## Test Steps

All unit tests pass.

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -607,6 +607,46 @@
     "@types/yargs" "^16.0.0"
     chalk "^4.0.0"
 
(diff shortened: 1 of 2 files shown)


mprew97 (approved):
👍

PR #155, opened 2022-09-14
[MOB-4840] Fix iframe height setting

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4840](https://iterable.atlassian.net/browse/MOB-4840)

## Description

The iframe document body has an 8px margin all around. The iframe height is set based on the body's scrollheight without taking the margin into consideration. Adds 16px to the iframe's scrollheight to eliminate the scrollbar that occurs when user uses the WYSIWYG editor to create their in-app.

## Test Steps

1. Create WYSIWYG in-app template
2. Send to yourself
3. Render in web sample app
4. No scrollbar

### DIFF
--- a/src/inapp/utils.ts
+++ b/src/inapp/utils.ts
@@ -423,8 +423,10 @@ export const paintIFrame = (
         };
 
         if (position !== 'Full') {
+          const iframeHeight =
+            iframe.contentWindow?.document?.body?.scrollHeight;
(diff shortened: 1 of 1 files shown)


ts-nguyen:
Could you add a quick comment for context just so it doesn't get regressed later on accidentally?

ts-nguyen (approved):
LGTM

PR #161, opened 2022-09-29
[MOB-4490] Deprecate boolean from getInAppMessages showInAppMessagesAutomatically param

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4490](https://iterable.atlassian.net/browse/MOB-4490)

## Description

getInAppMessages has a second argument which can look like this:

```
getMessages(...options, true);

getMessages(...options, { display: string; })
```

We want to deprecate the usage of the first method. The second method is the newer, scalable option which allows customers to defer the display of the in-app messages to a time later in the future. The boolean value was the original implementation before this option was introduced.

The reason it still exists in the codebase is for backwards compatibility purposes.

## Test Steps

All unit tests pass.

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -66,16 +66,14 @@ export function getInAppMessages(
 };
 export function getInAppMessages(
   payload: InAppMessagesRequestParams,
-  showInAppMessagesAutomatically?:
-    | boolean
-    | { display: 'immediate' | 'deferred' }
+  showInAppMessagesAutomatically?: { display: 'immediate' | 'deferred' }
 ) {
   clearMessages();
   const dupedPayload = { ...payload };
 
   /* delete SDK-defined payload props and email and userId */
-  delete (dupedPayload as any).userId;
-  delete (dupedPayload as any).email;
+  delete (dupedPayload as any).userId; // eslint-disable-line @typescript-eslint/no-explicit-any
+  delete (dupedPayload as any).email; // eslint-disable-line @typescript-eslint/no-explicit-any
   delete dupedPayload.displayInterval;
   delete dupedPayload.onOpenScreenReaderMessage;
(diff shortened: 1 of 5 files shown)


src/inapp/inapp.ts
```diff
@@ -74,16 +66,14 @@ export function getInAppMessages(
...
+  showInAppMessagesAutomatically?: { display: 'immediate' | 'deferred' }
 ) {
   clearMessages();
   const dupedPayload = { ...payload };
 
   /* delete SDK-defined payload props and email and userId */
-  delete (dupedPayload as any).userId;
-  delete (dupedPayload as any).email;
+  delete (dupedPayload as any).userId; // eslint-disable-line @typescript-eslint/no-explicit-any
+  delete (dupedPayload as any).email; // eslint-disable-line @typescript-eslint/no-explicit-any
```
pauljung14, line 76: QoL update to remove no-explicit-any warning from compilation logs

src/inapp/inapp.ts
```diff
@@ -53,14 +53,6 @@ export const clearMessages = () => {
...
-  payload: InAppMessagesRequestParams,
-  showInAppMessagesAutomatically: true
-): {
-  pauseMessageStream: () => void;
-  resumeMessageStream: () => Promise<HTMLIFrameElement | ''>;
-  request: () => IterablePromise<InAppMessageResponse>;
-};
 export function getInAppMessages(
   payload: InAppMessagesRequestParams,
   showInAppMessagesAutomatically: { display: 'immediate' | 'deferred' }
```
mprew97, line 58: kind of a nit, but can we type this? 

```
enum DISPLAY = {
  immediate = 'immediate',
  deffered = 'deferred',
}
```

mprew97 (approved):

PR #163, opened 2022-10-04
[MOB-4055] Update getMessages requests to use new web endpoint

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4055](https://iterable.atlassian.net/browse/MOB-4055)

## Test Steps

### DIFF
--- a/react-example/cypress/integration/inapp/inapp.spec.ts
+++ b/react-example/cypress/integration/inapp/inapp.spec.ts
@@ -5,7 +5,7 @@ describe('Requesting In-App Messages', () => {
     cy.intercept(
       {
         method: 'GET',
-        url: '/api/inApp/getMessages*'
+        url: '/api/inApp/web/getMessages*'
       },
       { fixture: 'inapp/200.json' }
     ).as('getInAppMessages');
@@ -28,7 +28,7 @@ describe('Requesting In-App Messages', () => {
     cy.intercept(
       {
         method: 'GET',
-        url: '/api/inApp/getMessages*'
+        url: '/api/inApp/web/getMessages*'
       },
       { fixture: 'inapp/400.json', statusCode: 400 }
     ).as('getInAppMessages');
@@ -55,7 +55,7 @@ describe('Auto-Display In-App Messages', () => {
     cy.intercept(
       {
         method: 'GET',
-        url: '/api/inApp/getMessages*'
+        url: '/api/inApp/web/getMessages*'
       },
       { fixture: 'inapp/200.json' }
     ).as('getInAppMessages');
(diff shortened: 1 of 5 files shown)


react-example/cypress/integration/inapp/inapp.spec.ts
```diff
@@ -5,7 +5,7 @@ describe('Requesting In-App Messages', () => {
     cy.intercept(
       {
         method: 'GET',
-        url: '/api/inApp/getMessages*'
+        url: '/api/inApp/web/getMessages*'
```
mprew97, line 8: can we turn this into a variable? that way, if it ever changes again we don't need to update it in multiple places :)

mprew97: in fact, we could probably make it some global const and use it in all of these places.

react-example/cypress/integration/inapp/inapp.spec.ts
```diff
@@ -1,11 +1,13 @@
 import * as mockMessages from '../../fixtures/inapp/200.json';
 
+const url = `/api/inApp/web/getMessages*`;
```
pauljung14, line 3: Opted to keep this as a local variable rather than reach outside of the react-example directory.

.env.example
```diff
@@ -1,5 +1,7 @@
 # Optional environment variables for local development. Start by creating a
-# called .env and add these values to it and change them appropriately
+# called .env and add these values to it and change them appropriately.
+# Remember to uncomment the variables!
 
 # BASE_URL="https://api.iterable.com/api"
+# GETMESSAGES_PATH="/inApp/web/getMessages"
```
pauljung14, line 6: Including in the "getting started" .env example file.

react-example/src/views/InApp.tsx
```diff
@@ -95,7 +95,7 @@ export const InApp: FC<{}> = () => {
   return (
     <>
       <h1>inApp Endpoints</h1>
-      <Heading>POST /inApp/getMessages</Heading>
+      <Heading>POST /inApp/web/getMessages</Heading>
```
pauljung14, line 98: This is also in the react-example app, minimal change to text that I think we can just keep.

src/constants.ts
```diff
@@ -6,6 +6,9 @@ export const RETRY_USER_ATTEMPTS = 0;
 
 export const BASE_URL = process.env.BASE_URL || 'https://api.iterable.com/api';
 
+export const GETMESSAGES_PATH =
+  process.env.GETMESSAGES_PATH ?? '/inApp/web/getMessages';
```
pauljung14, line 10: Will default to the web endpoint.

react-example/cypress/integration/inapp/inapp.spec.ts
pauljung14: Great suggestion, updated the PR!

mprew97 (approved):
🚢 :shipit:

PR #164, opened 2022-10-04
[MOB-4933] Delete expired messages from cache & only add new messages if there is space in cache

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4933](https://iterable.atlassian.net/browse/MOB-4933)

## Description

By default (as consistent with our other mobile SDKs), we persist messages in browser cache indefinitely until it reaches the expiration date. Automatically delete expired messages from the cache. Also, calculate remaining cache space and add new messages only if it fits in the cache.

## Test Steps

Send yourself a message that will quickly expire.
Call getMessages.
Check that all messages appear in cache (assuming there's space).
After message expires, call getMessages again.
Observe that expired message gets removed from the cache.

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -1,4 +1,4 @@
-import { delMany, entries, setMany } from 'idb-keyval';
+import { delMany, entries } from 'idb-keyval';
 import _set from 'lodash/set';
 import {
   ANIMATION_DURATION,
@@ -26,7 +26,9 @@ import {
 } from './types';
 import {
   addButtonAttrsToAnchorTag,
+  addNewMessagesToCache,
   addStyleSheet,
+  determineRemainingStorageQuota,
   filterHiddenInAppMessages,
   generateCloseButton,
   getHostnameFromUrl,
@@ -140,7 +142,7 @@ export function getInAppMessages(
 
       /** combine cached messages with NEW messages in delta response */
       const allMessages: Partial<InAppMessage>[] = [];
-      const newMessages: [string, InAppMessage][] = [];
+      const newMessages: { messageId: string; message: InAppMessage }[] = [];
       inAppMessages?.forEach((inAppMessage) => {
         /**
          * if message in response has no content property, then that means it is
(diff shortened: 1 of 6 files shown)


src/inapp/inapp.ts
```diff
@@ -1,16 +1,34 @@
-import { entries, set } from 'idb-keyval';
-import { throttle } from 'throttle-debounce';
```
pauljung14, line 2: Just reordering imports from here down

pauljung14, line 2: Just reordering imports from here down

src/inapp/inapp.ts
```diff
@@ -135,7 +143,8 @@ export function getInAppMessages(
 
       /** combine cached messages with NEW messages in delta response */
       const allMessages: Partial<InAppMessage>[] = [];
-      inAppMessages.forEach((inAppMessage) => {
+      const newMessages: { messageId: string; message: InAppMessage }[] = [];
```
pauljung14, line 146: Needed later when adding messages to the cache

src/inapp/utils.ts
```diff
@@ -114,6 +115,103 @@ export const sortInAppMessages = (messages: Partial<InAppMessage>[] = []) => {
   return messages.sort(by(['priorityLevel', 'asc'], ['createdAt', 'asc']));
 };
 
+/**
+ * detect amount of local storage remaining (quota) and used (usage).
+ * if usageDetails exist (not supported in Safari), use this instead of usage.
+ */
+export const determineRemainingStorageQuota = async () => {
```
pauljung14, line 122: @ts-nguyen @mprew97 I'm not quite happy with the implementation I've done for either of these utility methods because they feel a bit clunky and convoluted. Would love some feedback on these in particular

src/inapp/inapp.ts
```diff
@@ -153,25 +162,35 @@ export function getInAppMessages(
...
+            newMessages.push({
+              messageId: inAppMessage.messageId,
+              message: inAppMessage as InAppMessage
+            });
         }
       });
 
+      /** add new messages to the cache if they fit in the cache */
+      const remainingQuota = await determineRemainingStorageQuota();
+      await addNewMessagesToCache(newMessages, remainingQuota);
```
ts-nguyen, line 180: Do we use `remainingQuota` anywhere else? Any reason we can't call that util from within `addNewMessagesToCache` instead of passing it in?

src/inapp/utils.ts
```diff
@@ -114,6 +115,103 @@ export const sortInAppMessages = (messages: Partial<InAppMessage>[] = []) => {
...
+  try {
+    const storage: BrowserStorageEstimate | undefined =
+      'storage' in navigator && 'estimate' in navigator.storage
+        ? await navigator.storage.estimate()
+        : undefined;
+
+    /** 50 MB is the typical web browser cache quota for mobile devices */
+    const mobileBrowserQuota = 52428800;
+    /** max quota of browser storage that in-apps will potentially fill */
+    const inAppMaxBrowserQuota = storage?.quota && storage.quota * 0.6;
```
ts-nguyen, line 132: Is the 0.6 of the quota documented somewhere as the max we should use or a little arbitrary with the initial release? Could we put some comments on how we got that number?

src/inapp/utils.ts
```diff
@@ -114,6 +115,103 @@ export const sortInAppMessages = (messages: Partial<InAppMessage>[] = []) => {
...
+    /** determine lower max quota that can be used for message cache */
+    const messageQuota =
+      inAppMaxBrowserQuota && inAppMaxBrowserQuota < mobileBrowserQuota
+        ? inAppMaxBrowserQuota
+        : mobileBrowserQuota;
+
+    /** how much local storage is being used */
+    const usage = storage?.usage && storage.usage;
+    const idbUsage =
+      storage?.usageDetails?.indexedDB && storage.usageDetails.indexedDB;
```
ts-nguyen, line 142: Do we need the `&&` check if we're optionally chaining?

ts-nguyen: Also since they have similar usage can we combine the assignment into one variable so we don't have to check below which one to use. `usage = storage?.usageDetails?.indexedDB ?? storage?.usage`

PR #168, opened 2022-10-12
[MOB-4718] Fix handleLinks support for iframe links in Safari

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4718](https://iterable.atlassian.net/browse/MOB-4718)

## Description

Safari blocks all bound event handlers (including our click event handlers) in iframes, so links will not work in Safari unless we circumvent the restriction by appending target to each link tag (only if it doesn't already have a target attribute).

## Test Steps
TBD (test cases for permutations of Safari and Chrome and target and handleLinks)

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -459,9 +459,7 @@ export function getInAppMessages(
 
           const ua = navigator.userAgent;
           const isSafari =
-            !!ua.match(/safari/i) &&
-            !ua.match(/chrome|chromium|crios/i) &&
-            !ua.match(/firefox|fxios/i);
+            !!ua.match(/safari/i) && !ua.match(/chrome|chromium|crios/i);
 
           /* now we'll add click tracking to _all_ anchor tags */
           const links =
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -457,12 +457,17 @@ export function getInAppMessages(
             Promise.all(trackRequests).catch((e) => e);
           }
 
+          const ua = navigator.userAgent;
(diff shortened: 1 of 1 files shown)


src/inapp/inapp.ts
```diff
@@ -457,12 +457,17 @@ export function getInAppMessages(
             Promise.all(trackRequests).catch((e) => e);
           }
 
+          const ua = navigator.userAgent;
+          const isSafari =
```
ts-nguyen, line 461: Do we need to check that the browser is not safari? Can a browser give back safari and chrome or firefox as the user agent?

ts-nguyen (approved):
Awesome! A lot cleaner than I expected too

src/inapp/inapp.ts
pauljung14: From my testing, I found that in Chrome the user agent looks like this `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36` so then Chrome would be mistakenly thought to be Safari.

pauljung14: I just checked Firefox and found that its user agent looks like this `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:105.0) Gecko/20100101 Firefox/105.0` so it seems that at least I don't need to check that it's not firefox

pauljung14: Also checked on Microsoft Edge, whose user agent includes "Chrome" so the Safari check covers that.

PR #169, opened 2022-10-13
[MOB-4955] Fix target attribute conditional + refactor handleLinks

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4955](https://iterable.atlassian.net/browse/MOB-4955)

## Test Steps
Test all link permutations (no target, target, no target with handleLinks, target with handleLinks) with Chrome, Firefox, and Safari

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -493,12 +493,12 @@ export function getInAppMessages(
               Safari blocks all bound event handlers (including our click event handlers)
               in iframes, so links will not work in Safari unless we circumvent the
               restriction by appending target to each link tag.
+
+              NOTE: Because click event handlers cannot be attached to iframe links in
+              Safari, we cannot track in-app clicks in metrics.
             */
             if (isSafari) {
-              if (
-                !isIterableKeywordLink &&
-                link.getAttribute('target') === 'undefined'
-              ) {
+              if (!isIterableKeywordLink) {
                 if (typeof handleLinks === 'string') {
                   if (
                     handleLinks === 'open-all-same-tab' ||
@@ -508,7 +508,7 @@ export function getInAppMessages(
                   } else {
                     link.setAttribute('target', '_blank');
                   }
-                } else {
+                } else if (link.getAttribute('target') === null) {
                   link.setAttribute('target', '_top');
                 }
               }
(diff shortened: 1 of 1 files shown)


src/inapp/inapp.ts
```diff
@@ -481,6 +481,35 @@ export function getInAppMessages(
...
+                  handleLinks === 'open-all-same-tab' ||
+                  (isInternalLink && handleLinks === 'external-new-tab')
+                ) {
+                  sameTabAction();
+                } else {
+                  newTabAction();
+                }
+              } else if (link.getAttribute('target') === null) {
+                sameTabAction();
+              }
```
pauljung14, line 510: This logic was repeated, so moving to a helper fn

src/inapp/inapp.ts
```diff
@@ -481,6 +481,35 @@ export function getInAppMessages(
...
+              according to that enum. So the way this works is:
+
+              1. If _open-all-same-tab, then open every link in the same tab
+              2. If _open-all-new-tab, open all in new tab
+              3. If _external-new-tab_, open internal links in same tab, otherwise new tab.
+
+              This was a fix to account for the fact that Bee editor templates force
+              target="_blank" on all links, so we gave this option as an escape hatch for that.
+            */
+            const manageHandleLinks = (
```
pauljung14, line 495: This logic was repeated, so moving to a helper fn

src/inapp/inapp.ts
```diff
@@ -481,6 +481,35 @@ export function getInAppMessages(
...
+              if (typeof handleLinks === 'string') {
+                if (
+                  handleLinks === 'open-all-same-tab' ||
+                  (isInternalLink && handleLinks === 'external-new-tab')
+                ) {
+                  sameTabAction();
+                } else {
+                  newTabAction();
+                }
+              } else if (link.getAttribute('target') === null) {
```
pauljung14, line 508: getAttribute returns null rather than undefined when absent, this is the fix for the conditional logic

pauljung14: also moves this below the handleLinks if statement, because handleLinks is supposed to override target values

ts-nguyen (approved):

PR #170, opened 2022-10-13
[MOB-4956] Refactor link handling

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4956](https://iterable.atlassian.net/browse/MOB-4956)

## Description

Moving logic not related to handleLinks outside of the manageHandleLinks method.
Also make update to handle _self target for links in Safari.

## Test Steps

Test all link permutations (no target, target, no target with handleLinks, target with handleLinks) with Chrome, Firefox, and Safari

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -505,8 +505,6 @@ export function getInAppMessages(
                 } else {
                   newTabAction();
                 }
-              } else if (link.getAttribute('target') === null) {
-                sameTabAction();
               }
             };
 
@@ -535,6 +533,9 @@ export function getInAppMessages(
                     link.setAttribute('rel', 'noopener noreferrer');
                   }
                 );
+                if (!handleLinks && link.getAttribute('target') === null) {
+                  link.setAttribute('target', '_top');
+                }
               }
             } else {
               link.addEventListener('click', (event) => {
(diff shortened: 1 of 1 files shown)


src/inapp/inapp.ts
```diff
@@ -535,6 +532,13 @@ export function getInAppMessages(
                     link.setAttribute('rel', 'noopener noreferrer');
                   }
                 );
+                const targetAttr = link.getAttribute('target');
+                if (
+                  !handleLinks &&
+                  (targetAttr === null || targetAttr === '_self')
```
pauljung14, line 538: Need to handle _self case for Safari

src/inapp/inapp.ts
```diff
@@ -613,15 +617,21 @@ export function getInAppMessages(
...
-                    /**
-                      Using target="_blank" without rel="noreferrer" and rel="noopener"
-                      makes the website vulnerable to window.opener API exploitation attacks
-                      
-                      @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#security_and_privacy
-                    */
-                    if (openInNewTab)
-                      global.open(clickedUrl, '_blank', 'noopener,noreferrer');
-                    else global.location.assign(clickedUrl);
+                    if (!handleLinks) {
```
pauljung14, line 620: Need to check for handleLinks here because if handleLinks causes links to open in new tab, then we don't want the current tab to make it to the last else clause

src/inapp/inapp.ts
```diff
@@ -505,8 +504,6 @@ export function getInAppMessages(
                 } else {
                   newTabAction();
                 }
-              } else if (link.getAttribute('target') === null) {
-                sameTabAction();
```
ts-nguyen, line 509: Is this just to respect browser default instead?

ts-nguyen (approved):
LGTM!

src/inapp/inapp.ts
pauljung14: Yep that's right. I decided to pull it out of the `manageHandleLinks` method because the conditions for non-handleLinks scenarios slightly differed between Safari and non-Safari.

PR #171, opened 2022-10-14
[MOB-4959] Do not add autogenerated close button if in-app is displayed in Safari

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4959](https://iterable.atlassian.net/browse/MOB-4959)

## Description

Autogenerated close button doesn't work on Safari (desktop and mobile)

## Test Steps
On Safari:
1. Create a web in-app message, including an auto-generated close button 
2. Send to self and visit the website on a mobile device and desktop
3. Web in-app message loads
4. No auto-generated close button appears

On Chrome/Firefox/Edge:
1. Using same message as above, send to self and visit website on a mobile device and desktop
3. Web in-app message loads
4. Clicking the auto-generated close button closes the message

### DIFF
--- a/src/inapp/inapp.ts
+++ b/src/inapp/inapp.ts
@@ -353,6 +353,15 @@ export function getInAppMessages(
           const isSafari =
             !!ua.match(/safari/i) && !ua.match(/chrome|chromium|crios/i);
 
+          /**
+           * We allow users to dismiss messages by clicking outside of the
+           * message not only when isRequiredToDismissMessage is not true
+           * but also when browser is detected to be Safari, regardless of
+           * whether isRequiredToDismissMessage is true. Safari blocks
+           * all bound event handlers and so we cannot execute Javascript
+           * to listen for click events. As such, we should not prevent users
+           * from being able to dismiss the message by clicking outside of it.
+           */
           if (!payload.closeButton?.isRequiredToDismissMessage || isSafari) {
             overlay.addEventListener('click', () => {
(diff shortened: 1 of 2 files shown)


src/inapp/inapp.ts
```diff
@@ -349,7 +349,11 @@ export function getInAppMessages(
             );
           }
 
-          if (!payload.closeButton?.isRequiredToDismissMessage) {
+          const ua = navigator.userAgent;
+          const isSafari =
+            !!ua.match(/safari/i) && !ua.match(/chrome|chromium|crios/i);
+
+          if (!payload.closeButton?.isRequiredToDismissMessage || isSafari) {
```
pauljung14, line 356: If browser is Safari, we do not render an auto-generated close button, because it requires JS to function. However, that means if the customer wants to require the close button to dismiss the message, we should override that feature by still allowing users to click outside the message to dismiss, otherwise there isn't a way to close it.

src/inapp/inapp.ts
```diff
@@ -420,7 +424,7 @@ export function getInAppMessages(
               values. This button is just a quality-of-life feature so that the customer will
               have an easy way to close the modal outside of the other methods.
             */
-            if (payload.closeButton) {
+            if (payload.closeButton && !isSafari) {
```
pauljung14, line 427: Do not show close button if Safari because it won't work (Safari blocks JS on bound event handlers)

mprew97 (approved):
dang safari!

PR #172, opened 2022-10-18
[MOB-4989] Render document string instead of iframe in sample app content html

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4989](https://iterable.atlassian.net/browse/MOB-4989)

## Description

When customer does not use the showInAppMessagesAutomatically param, then the return value for the request promise’s content.html is an iframe (same as if the customer used the triggerDisplayMessages method). For the rendering purposes of the sample app, use the request() method to provide the response with InAppMessage in the content.html response.

## Test Steps

Send html content message to self.
In sample app, click "Get messages (do not auto-display)".
Content.html value should be stringified html (for html content messages).

### DIFF
--- a/src/inapp/utils.ts
+++ b/src/inapp/utils.ts
@@ -370,7 +370,7 @@ const generateSecuredIFrame = () => {
 export const wrapWithIFrame = (html: string): HTMLIFrameElement => {
   const iframe = generateSecuredIFrame();
   iframe.onload = () => {
-    iframe.contentWindow?.document.write(html);
+    iframe.contentDocument?.write(html);
   };
   return iframe;
 };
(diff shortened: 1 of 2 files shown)


src/inapp/utils.ts
```diff
@@ -370,7 +370,7 @@ const generateSecuredIFrame = () => {
 export const wrapWithIFrame = (html: string): HTMLIFrameElement => {
   const iframe = generateSecuredIFrame();
   iframe.onload = () => {
-    iframe.contentWindow?.document.write(html);
+    iframe.contentDocument?.write(html);
```
pauljung14, line 373: unrelated oppt to refactor

mprew97 (approved):

PR #173, opened 2022-10-18
[MOB-4992] Support configuration option for using local cache

pauljung14:
## JIRA Ticket(s) if any

* [MOB-4992](https://iterable.atlassian.net/browse/MOB-4992)

## Description

We are adding a new API endpoint for message delta (to be used in conjunction with the clientside message caching feature) in the near future. For an improved compatibility experience, we will provide support for an optional `useLocalCache` parameter at a later time, but for GA launch this parameter will not be available to be set. Customers may choose to use this parameter when it becomes available with a minor version update, but customers who do not want/need this feature will be able to do fine without it.

## Test Steps

Send in-app messages to self.
Check that no `latestCachedMessageId` parameter exists on request.
No messages are added to the cache.

### DIFF
--- a/src/inapp/types.ts
+++ b/src/inapp/types.ts
@@ -1,3 +1,5 @@
+import { IterablePromise } from 'src/types';
+
 interface SDKInAppMessagesParams {
   displayInterval?: number;
   /* what should the screen reader say once the message opens */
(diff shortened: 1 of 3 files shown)


src/inapp/types.ts
```diff
@@ -37,6 +39,15 @@ export interface InAppMessagesRequestParams extends SDKInAppMessagesParams {
...
 
+export interface GetInAppMessagesResponse {
+  pauseMessageStream: () => void;
+  resumeMessageStream: () => Promise<HTMLIFrameElement | ''>;
+  request: () => IterablePromise<InAppMessageResponse>;
+  triggerDisplayMessages: (
+    messages: Partial<InAppMessage>[]
+  ) => Promise<HTMLIFrameElement | ''>;
+}
+
```
mprew97, line 50: love the typing here :)

mprew97 (approved):
🚢 :shipit:

PR #196, opened 2023-03-28
Fix safari close issues

mprew97:
## JIRA Ticket(s) if any

* [MOB-5930](https://iterable.atlassian.net/browse/MOB-5930)

## Description
Most browsers, if an i-frame has `allow-same-origin`, allow script execution by default. Older browsers and Safari do not. If we are going to `allow-same-origin` script execution on new browsers, then there shouldn't be any reason not to allow it on Safari. 

See https://caniuse.com/mdn-html_elements_iframe_sandbox-allow-same-origin
## Test Steps

[MOB-5930]: https://iterable.atlassian.net/browse/MOB-5930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/inapp/utils.ts
+++ b/src/inapp/utils.ts
@@ -4,7 +4,7 @@ import { ANIMATION_DURATION } from 'src/constants';
 import { WebInAppDisplaySettings } from 'src/inapp';
 import { srSpeak } from 'src/utils/srSpeak';
 import { trackInAppDelivery } from '../events';
-import { BrowserStorageEstimate, InAppMessage } from './types';
+import { BrowserStorageEstimate, CachedMessage, InAppMessage } from './types';
 
 interface Breakpoints {
   smMatches: boolean;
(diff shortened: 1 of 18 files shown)


src/inapp/utils.ts
```diff
@@ -359,12 +359,11 @@ const mediaQueryXl = global?.matchMedia?.('(min-width: 1301px)');
...
-  // event handlers on elements in it preventing our custom link handling
-  iframe.setAttribute(
-    'sandbox',
-    'allow-same-origin allow-popups allow-top-navigation'
-  );
+  /**
+   * Most other browsers, if "allow-same-origin" enabled, already "allow-scripts". Safari does not by default, and same with some older browsers so
+   * we must manually set it here. See https://caniuse.com/mdn-html_elements_iframe_sandbox-allow-same-origin for more information.
+   */
+  iframe.setAttribute('sandbox', 'allow-same-origin allow-scripts');
```
ts-nguyen, line 366: I think it's highly advised against to allow-scripts with allow-same-origin here, MDN docs for iframe says:

> When the embedded document has the same origin as the embedding page, it is strongly discouraged to use both allow-scripts and allow-same-origin, as that lets the embedded document remove the sandbox attribute — making it no more secure than not using the sandbox attribute at all.

We had to jump through a bunch of hoops for a security audit around JS execution so I could see adding this in being a problem. Unfortunately that was the reason we had to gate a bunch of functionality around Safari ☹️

mprew97: @ts-nguyen `allow-same-origin` allows for JS execution on newer browsers so if that's a security issue on Safari, then it's an issue with our non-safari implementation as well. If you look here https://caniuse.com/mdn-html_elements_iframe_sandbox-allow-same-origin , you'll see that other browsers allow script execution by default.

PR #199, opened 2023-04-04
[MOB-5933]: EU Domain Configuration

mprew97:
## JIRA Ticket(s) if any

* [MOB-5933](https://iterable.atlassian.net/browse/MOB-5933)

## Description

Adds env variable to allow users to set which domain (US or EU) they'd prefer to hit.

Opted to go for env var as opposed to a config setting since
a) that's how the URL is already being configured
b) this will allow customers to set their targets per their environment, allowing for more flexibility

## Test Steps

[MOB-5933]: https://iterable.atlassian.net/browse/MOB-5933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/constants.ts
```diff
@@ -4,7 +4,17 @@ export const DISPLAY_INTERVAL_DEFAULT = 30000;
...
+
+const US_ITERABLE_DOMAIN = 'api.iterable.com';
+
+const EU_ITERABLE_DOMAIN = 'eu.iterable.com';
+
+const ITERABLE_API_URL = `https://${
+  IS_EU_SERVICE ? EU_ITERABLE_DOMAIN : US_ITERABLE_DOMAIN
+}/api`;
+
+export const BASE_URL = process.env.BASE_URL || ITERABLE_API_URL;
```
pauljung14, line 17: @mprew97 Even if `IS_EU_SERVICE` is true, the `BASE_URL` env variable here will be used and the EU domain will just be the fallback.

What is the upside of allowing users to set an `IS_EU_SERVICE` variable as well as a `BASE_URL` variable? Would it suffice to just allow users to set `BASE_URL` in the env file to "https://eu.iterable.com/api" (or the US domain depending on their use)?

mprew97: To your first point, I would say if they want a custom URL, whether for development or whatever, then that custom URL should take priority. Maybe that indicates `BASE_URL` should be `CUSTOM_URL`.

To the second, `BASE_URL` could be a fallback instead if we don't anticipate users ever having custom URLs. It's really based on whatever is a better developer experience. I feel that at least just allowing them to set flags for their preferred env is a better experience since devs don't have to worry about typos/wrong urls, they can just toggle a flag for their preferred region. I'm also not sure if there's a need to specify a subdomain since all of the endpoints hit the same subdomain `api.iterable.com` for the time being. If that ever changes, we could extend it. 

Thoughts? @pauljung14

mprew97: @pauljung14 and I had a conversation around this behavior. Since `BASE_URL` still currently takes precedence, in order to make sure there are no breaking changes, we are going to leave the env var in, change the flag name to `IS_ITERABLE_EU_SERVICE`, and leave some comments about the priority of the environment variables.

ts-nguyen (approved):

PR #212, opened 2023-07-07
notificiation view implemented

iamavishkar:
## JIRA Ticket(s) if any

## Test Steps


Sagarpatel14:
image
image

src/components/notification/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    border: "none",
+    color: "#2196f3",
+    cursor: "pointer",
+  };
+
+  return (
+    <div style={cardStyle}>
+      <h2>{title}</h2>
+      <p>{description}</p>
+      {primaryButtonLabel ? (
```
mprew97, line 53: nit: use `&&` here and remove the need for `: null`

src/components/notification/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <button
+          onClick={onClickPrimaryBtn}
+          style={{ ...primaryButtonDefaultStyle, ...primaryButtonStyle }}
+        >
+          {primaryButtonLabel}
+        </button>
+      ) : null}
+      {secondaryButtonLabel ? (
```
mprew97, line 61: nit: same as above

iamavishkar:
request you to please review the PR as it has been updated.

mprew97:

Ty sir!

We should be able to just import the `Notification` component in `react-example` as a user would. Can we do that and also have an example of it being displayed on a page somewhere in the example? Like an actual live component example?

devcsomnicg:
Hi @mprew97 here is the video for the same: https://drive.google.com/file/d/1ifip1OplImM9m2q-kOXGwxNwenqJXkCC/view?usp=sharing

PR #213, opened 2023-07-07
banner view implemented

iamavishkar:
## JIRA Ticket(s) if any

## Test Steps


Sagarpatel14:
image
image

src/components/banner/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+              order: 2;
+              max-width:75%;
+            }
+            .banner-container > div:last-child {
+              order: 1;
+              margin-bottom: 0;
+            }
+          }
+        `}
+      </style>
```
mprew97, line 106: i'd really like to find a decent library to implement granular css styles like this. can this not be passed in directly to the `style` props same as the others? As much as possible, would prefer to pull this out of the actual component tree.

mprew97 (changes requested):
We need to pull this into a new page in the react example app `/components` and render it there so it can be seen/used in real time.

iamavishkar:
request you to please review the PR as it has been updated.

mprew97:

Ty sir!

We should be able to just import the `Banner` component in `react-example` as a user would, as opposed to duplicating it. Can we do that and also have an example of it being displayed on a page somewhere in the example? Like an actual live component example?

devcsomnicg:
Hi @mprew97 here is the video for the same: https://drive.google.com/file/d/1ifip1OplImM9m2q-kOXGwxNwenqJXkCC/view?usp=sharing

mprew97 (approved):

mprew97:
Once the build issues are resolved, we can merge this.

mprew97:
Hi @devcsomnicg can we resolve the conflict here?

PR #214, opened 2023-07-07
card view implemented

iamavishkar:
## JIRA Ticket(s) if any

## Test Steps


Sagarpatel14:
image
image

src/components/card/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  onClickSecondaryBtn?: () => void;
+  imgStyle?: CSSProperties;
+  titleStyle?: CSSProperties;
+  cardStyle?: CSSProperties;
+  textStyle?: CSSProperties;
+  primaryBtnStyle?: CSSProperties;
+  secondaryBtnStyle?: CSSProperties;
+}
+
+const Card = (props: CardProps) => {
```
mprew97, line 22: would prefer to keep this consistent with the other components and destructure the props either inline in the component declaration or just below.

src/components/card/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const isPrimaryBtnDisabled = props.disablePrimaryBtn || false;
```
mprew97, line 23: these should both just be `Boolean(someProp)`

src/components/card/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+            .card {
+              max-width: 90%;
+            }
+          }
+          .card-buttons {
+            display: flex;
+            margin-top: 40px;
+            margin-bottom: 10px;
+            margin-right: 50px;
+          }`}
```
mprew97, line 120: Same comment as the other PR: it's preferred to keep styles out of the component tree as much as possible. if the `style` prop doesn't allow certain css in `style`, we should consider other options.

mprew97 (changes requested):
Please add this along with the other components to a `/components` page in the react-example app.

mprew97:

Ty sir!

We should be able to just import the `Card` component in `react-example` as a user would. Can we do that and also have an example of it being displayed on a page somewhere in the example? Like an actual live component example?

devcsomnicg:
Hi @mprew97 here is the video for the same:

https://drive.google.com/file/d/1ifip1OplImM9m2q-kOXGwxNwenqJXkCC/view?usp=sharing

mprew97:
Please resolve the build issues here and then this can be merged.

mprew97:
@devcsomnicg can we resolved the conflict here? happy to merge as soon as it's resolved.

PR #223, opened 2023-07-20
Firebase/messaging

devcsomnicg:
## JIRA Ticket(s) if any

## Test Steps


mprew97 (commented):
.

src/message/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      }
+    });
+  };
+  
+  private async registerBrowserToken (token: string) {
+    return baseIterableRequest<IterableResponse>({
+      method: 'POST',
+      url: REGISTER_TOKEN,
+      headers: {
+        "Api-Key": "1d925bbf54f34bcb92a15da14bfb5d1d"
```
mprew97, line 63: Need to remove the hardcoded api key here from the commit history and make this use the environment variable in `.env.example`.

devcsomnicg: Ok

devcsomnicg: Actually, we added the "initialize(process.env.API_KEY)" call on the example app, to add "API_KEY" for API calls, that's why we removed the static header

PR #242, opened 2023-11-02
Dev 248 (Update payload to include updated placements)

devcsomnicg:
## Description

- **Release Notes**
> Update payload to include updated placements

> Update the sample app

- **Test Steps**
> To start, access the sample app code and navigate to react-example > src > views > Embedded.tsx file. Modify the placement ID you wish to test. Run the code, select the 'embedded' option, and input your user ID. Then, click the 'Fetch Embedded Messages' button to test and view the embedded messaging list for that specific placement ID.


src/embedded/embeddedManager.ts
```diff
@@ -14,16 +14,44 @@ export class EmbeddedManager {
...
+    packageName: string,
+    placementIds: number[]
+  ) {
     try {
+      let url = `${embedded_msg_endpoint}?userId=${userId}`;
+      url += `&email=${email}`;
+      url += `&platform=${platform}`;
+      url += `&sdkVersion=${sdkVersion}`;
+      url += `&packageName=${packageName}`;
+      url += placementIds.map((id) => `&placementIds=${id}`).join('');
```
mprew97, line 51: placementIds should be made optional. A customer should have the option to pass _no_ placement ids. if there are no ids, we shouldn't append this string to the `url` string.

mprew97 (changes requested):

src/embedded/embeddedManager.ts
devcsomnicg: Done

PR #252, opened 2023-12-23
Create test (#250)

hani-iterable:
## JIRA Ticket

* [MOB-7534](https://iterable.atlassian.net/browse/MOB-7534)

## Description

- **Release Notes**
> The required changes for correct email/userId field placement have been completed

> We've also updated the sample app to showcase the changes.

- **Test Steps**
> Start by launching the sample app, then select the 'embedded' option. Enter your userId/email in the designated field. Afterward, choose the option you wish to test and click submit button. You will then be able to observe the parameter changes in the API request.

[MOB-7534]: https://iterable.atlassian.net/browse/MOB-7534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react-example/src/components/EmbeddedForm.tsx
```diff
@@ -102,12 +103,13 @@ export const EmbeddedForm: FC<Props> = ({
     const appPackageName = 'my-lil-site';
 
     trackEmbeddedMessageClick(
-      userId,
       payload,
       buttonIdentifier,
       clickedUrl,
       appPackageName,
-      1627060811283
+      1627060811283,
```
mprew97, line 110: should this actually be hardcoded?

react-example/src/components/EmbeddedForm.tsx
```diff
@@ -154,7 +156,8 @@ export const EmbeddedForm: FC<Props> = ({
     setTrackingEvent(true);
 
     const sessionData = {
-      userId: userId,
+      userId: emailRegex.test(userId) ? undefined : userId,
```
mprew97, line 159: you should be able to just pass in the value but determine which key to use dynamically. would also suggest making a helper/util method to do so.

mprew97: See https://www.freecodecamp.org/news/how-to-set-dynamic-object-properties-using-computed-property-names/

mprew97 (changes requested):

react-example/src/components/EmbeddedForm.tsx
hani-iterable: I have changed to actual value.

react-example/src/components/EmbeddedForm.tsx
hani-iterable: Modified both the SDK and the sample app to incorporate support for 'Computed Property Names' wherever applicable.

mprew97 (approved):

PR #283, opened 2023-12-26
Embedded text wrap

hani-iterable:
## JIRA Ticket

* [MOB-7514](https://iterable.atlassian.net/browse/MOB-7514)

## Description

> Modify the style to resolve text wrapping issue.

## Test Steps

> Start by launching the sample app, then select the 'embedded Msgs' option. Enter your userId/email in the designated field. Afterward, choose the option you wish to test and see the result.

[MOB-7514]: https://iterable.atlassian.net/browse/MOB-7514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/card/index.tsx
```diff
@@ -87,15 +92,28 @@ export const Card = (props: ICardProps) => {
...
         {primaryBtnLabel ? (
           <button
             disabled={disablePrimaryBtn}
-            style={{ ...defaultButtonStyles, ...primaryBtnStyle }}
+            style={
+              disablePrimaryBtn
+                ? {
+                    ...defaultButtonStyles,
+                    ...primaryBtnStyle,
+                    color: 'grey'
```
mprew97, line 113: Think we should allow users to pass their own color here while we maintain the default `grey`.

mprew97 (changes requested):
Going forward, we should try and not have branches based on previous branches when their changes aren't necessarily dependent on each other. This makes it hard to determine what is actually just meant to be in this PR and what is meant to be in another.

src/components/card/index.tsx
hani-iterable: set different style for disabled button.

mprew97 (approved):

mprew97 (approved):

PR #301, opened 2024-02-08
Integrate AUT feature

hani-iterable:


src/utils/anonymousUserMerge.ts
```diff
@@ -0,0 +1,111 @@
...
+  ENDPOINT_MERGE_USER
+} from 'src/constants';
+import { AnonymousUserEventManager } from './anonymousUserEventManager';
+import { baseIterableRequest } from '../request';
+import { IterableResponse } from '../types';
+
+export class AnonymousUserMerge {
+  private anonymousUserManager = new AnonymousUserEventManager();
+
+  mergeUserUsingUserId(destinationUserId: string): void {
```
mprew97, line 15: mergeUserUsingUserId and mergeUserUsingEmail should be joined. we can determine whether to use one or the other in the method itself.

src/utils/anonymousUserMerge.ts
```diff
@@ -0,0 +1,111 @@
...
+          if (dataObj.user) {
+            this.callMergeApi(
+              '',
+              sourceUserId,
+              localStorage.getItem(SHARED_PREF_EMAIL) || '',
+              destinationUserId
+            );
+          }
+        }
+      })
```
mprew97, line 44: do we not need to handle for the case where userData does not exist?

src/utils/anonymousUserMerge.ts
```diff
@@ -0,0 +1,111 @@
...
+    baseIterableRequest<IterableResponse>({
+      method: 'POST',
+      url: ENDPOINT_MERGE_USER,
+      data: {
+        sourceEmail: sourceEmail,
+        sourceUserId: sourceUserId,
+        destinationEmail: destinationEmail,
+        destinationUserId: destinationUserId
+      },
+      validation: {}
```
mprew97, line 100: why is the `validation` field hardcoded?

src/utils/anonymousUserMerge.ts
```diff
@@ -0,0 +1,111 @@
...
+    })
+      .then((response) => {
+        if (response.statusText === 'success') {
```
mprew97, line 103: not sure if the SDK already has it, but a constant for the `statusText` values would be a nice to have.

mprew97 (changes requested):
I know this is. WIP but would like to get a jump on the review here. 

A few things:

- We need to remove the `package-lock` file as our team uses yarn.
- We need to test the hell out of all of the new additional `utils` files. That is the most complext portion of this work and they currently have no testing around them. 
- Going forward, I'd like to see us break up these features into smaller PRs. As an example, in this PR alone, we could have a PR for the AUT base logic, a PR for the new types, a PR for building this out in the react-example app, etc. This makes it easier for us to review and get the work merged and tested.

mprew97:
One more thing, please point this branch to `AUT` as we will want to test functionality there. Any additional changes can be pushed to that branch until we are ready to go live with this.

src/events/events.ts
```diff
@@ -3,12 +3,23 @@ import { InAppEventRequestParams, InAppTrackRequestParams } from './types';
...
-  delete (payload as any).userId;
-  delete (payload as any).email;
-
+  if (
+    (!('userId' in payload) ||
+      payload.userId === null ||
+      typeof payload.userId === 'undefined') &&
+    (!('email' in payload) ||
+      payload.email === null ||
+      typeof payload.email === 'undefined')
```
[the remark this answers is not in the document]
hani-iterable: Done

src/commerce/commerce.ts
```diff
@@ -27,12 +38,22 @@ export const updateCart = (payload: UpdateCartRequestParams) => {
...
+    (!payload.user ||
+      !('email' in payload.user) ||
+      payload.user.email === null ||
+      typeof payload.user.email === undefined)
+  ) {
+    const anonymousUserEventManager = new AnonymousUserEventManager();
+    anonymousUserEventManager.trackAnonPurchaseEvent(payload);
+    const errorMessage =
+      'Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods';
+    throw new Error(errorMessage);
```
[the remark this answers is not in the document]
hani-iterable: The method is structured in a way that doesn't allow for direct returns.

src/events/events.test.ts
```diff
@@ -244,95 +271,97 @@ describe('Events Requests', () => {
...
-      messageId: 'fdsafd',
-      deviceInfo: { appPackageName: 'my-lil-site' }
-    } as any);
-    const trackOpenResponse = await trackInAppOpen({
-      email: 'hello@gmail.com',
-      userId: '1234',
-      messageId: 'fdsafd',
-      deviceInfo: { appPackageName: 'my-lil-site' }
-    } as any);
+  // it('should not send up passed email or userId params', async () => {
```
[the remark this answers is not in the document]
hani-iterable: Resolved

src/events/events.ts
```diff
@@ -77,8 +88,8 @@ export const trackInAppClick = (
   sendBeacon = false
 ) => {
   /* a customer could potentially send these up if they're not using TypeScript */
-  delete (payload as any).userId;
-  delete (payload as any).email;
+  // delete (payload as any).userId;
+  // delete (payload as any).email;
```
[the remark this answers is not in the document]
hani-iterable: Just for testing purpose. I have undo that.

src/users/users.ts
```diff
@@ -20,10 +21,20 @@ export const updateUserEmail = (newEmail: string) => {
...
+      typeof payload.userId === 'undefined') &&
+    (!('email' in payload) ||
+      payload.email === null ||
+      typeof payload.email === 'undefined')
+  ) {
+    const anonymousUserEventManager = new AnonymousUserEventManager();
+    anonymousUserEventManager.trackAnonUpdateUser(payload);
+    const errorMessage =
+      'Iterable SDK must be initialized with an API key and user email/userId before calling SDK methods';
+    throw new Error(errorMessage);
```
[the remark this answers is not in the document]
hani-iterable: Done

src/utils/anonymousUserMerge.ts
```diff
@@ -0,0 +1,111 @@
...
+  private anonymousUserManager = new AnonymousUserEventManager();
+
+  mergeUserUsingUserId(destinationUserId: string): void {
+    const sourceUserId = localStorage.getItem(SHARED_PREF_USER_ID);
+    if (
+      !sourceUserId ||
+      sourceUserId === '' ||
+      sourceUserId === destinationUserId
+    ) {
+      console.log('sourceUserId is null or same as destinationUserId');
```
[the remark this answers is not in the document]
hani-iterable: Done

src/utils/anonymousUserMerge.ts
hani-iterable: We need both methods, as we need to call separate APIs for each and it will make method complex to put many checks.

src/utils/anonymousUserMerge.ts
hani-iterable: No. In the requirement that if API provides success response then and only then we have to do further operations.

src/utils/anonymousUserMerge.ts
hani-iterable: The code error due to copy-pasting has been rectified.

src/utils/anonymousUserMerge.ts
hani-iterable: Done

src/utils/anonymousUserMerge.ts
mprew97: Here is an [example](https://github.com/Iterable/iterable-web-sdk/pull/304) of how this could be cleaned up. This hasn't been tested yet and the unit tests haven't been changed but should give you an idea.

Also, we should consider returning something from that call as opposed to `void`. I imagine devs would want to know whether it succeeded or failed.

mprew97: Another thought, we should also pass in the `source`, so the `mergeUser` function takes in `destination` and `user` so we don't have that weird early return statement. That, way, the method is only concerned about making the proper api calls, not pulling data from local storage. If there's no local storage data or the target equals the source, then we shouldn't call the method to begin with. Those checks should happen higher up in the call order.

mprew97: So really, the `mergeUser` method should take in this type as it's param, same as the `callMergeApi` in my example and the logic to get those params should happen outside the scope of this method somewhere. 

```
  type MergeApiParams = {
  sourceEmail: string | null;
  sourceUserId: string | null;
  destinationEmail: string | null;
  destinationUserId: string | null;
}
```

hani-iterable: Thank you for the new vision for the development. I have implemented that way and also modified the test cases accordingly.

mprew97 (changes requested):
Can we fix the build issue and remove the additional `package-lock` file? we ought to continue using `yarn` for our package manager.

src/utils/anonymousUserEventManager.ts
```diff
@@ -0,0 +1,368 @@
...
+      const userSessionInfo = JSON.parse(userData);
+      const userDataJson = userSessionInfo[SHARED_PREFS_ANON_SESSIONS];
+      const payload: TrackAnonSessionParams = {
+        email: this.getEmail() || undefined,
+        userId: this.getEmail() ? null : this.getUserID() || undefined,
+        createdAt: Math.round(new Date().getTime() / 1000) | 0,
+        deviceInfo: {
+          appPackageName: '',
+          deviceId: '',
+          platform: 'web'
```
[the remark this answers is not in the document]
hani-iterable: I have passed the value following the same pattern as other methods in our codebase.

PR #302, opened 2024-02-14
[MOB-7613] Remove extra height from iframe + refactor

pauljung14:
## JIRA Ticket(s) if any

* [MOB-7613](https://iterable.atlassian.net/browse/MOB-7613)

## Description

Remove extra height added to iframe style. Originally addressed styling issue with WYSIWYG templates, but is handled better now by simply overriding the browser's user agent stylesheet on the body element itself.

Set the iframe height on load, because the image preloader is not a foolproof way to ensure all images get loaded in advance (I tested this to be the case for  tags).

Refactor cache-related utils to separate file.
Refactor types.

## Test Steps

Given I send myself templates made with BEE and WYSIWYG editors
When I render the messages on my browser
Then I see no unexpected margin around the message
And I see the iframe height is automatically adjusted based on all images loaded in the message

Regression testing on general functionality, esp. for...
* display position (center, top right, bottom right, full)
* display option (immediate vs deferred)
* close button position (top right, top left)
* link handling (open new tab, open same tab, external new tab)
...because typings have been refactored for these.

[MOB-7613]: https://iterable.atlassian.net/browse/MOB-7613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react-example/.env.example
pauljung14, line 1: Convenience variable in our sample app, because I'm always setting the jwt generator url to something else in the index file anyway (and I figure customers who try out our sample app would probably use a different jwt generator than setting one up on their localhost at that exact port).

src/inapp/utils.ts
```diff
@@ -445,6 +447,16 @@ export const wrapWithIFrame = (html: string): HTMLIFrameElement => {
   return iframe;
 };
 
+/**
+ * Default Chrome user agent stylesheet adds 8px margin to body element.
+ * This unsets it so that it does not interfere with iframe sizing.
+ */
```
pauljung14, line 453: Screenshot 2024-02-15 at 10 14 07 AM

src/inapp/utils.ts
```diff
@@ -591,27 +621,27 @@ export const paintIFrame = (
...
-            iframe.contentWindow?.document?.body?.scrollHeight;
-          /*
-            For web in-app messages created with WYSIWYG editor,
-            there is 8px margin all around the iframe document body.
-            Add 16px to the iframe height to eliminate scrollbar.
-            Add 10px to eliminate discrepancy between the height
-            of the document html and that of the iframe itself.
-          */
-          iframe.style.height =
-            ((iframeHeight && iframeHeight + 26) || 0) + 'px';
```
pauljung14, line 604: This is now handled by unsetting the margin set by the browser's default user agent stylesheet

src/inapp/tests/cache.test.ts
pauljung14, line 1: Copy pasted from `src/inapp/request.ts`

src/inapp/cache.ts
pauljung14, line 1: Copy pasted from `src/inapp/utils.ts`

pauljung14: One small step to wrangling the gigantic utils file

pauljung14: None of it is being used atm (as you can see from `inapp.ts`), because we were waiting for BE support for a new getMessages endpoint with cache support

src/inapp/utils.ts
```diff
@@ -591,27 +498,27 @@ export const paintIFrame = (
...
     if (srMessage) {
       srSpeak(srMessage, 'assertive');
     }
+
+    iframe.onload = () => {
+      if (position !== 'Full') {
+        const scrollHeight = iframe.contentDocument?.body.scrollHeight ?? 0;
+        if (scrollHeight) iframe.style.height = scrollHeight + 'px';
+      }
+    };
```
pauljung14, line 520: This is a failsafe for img tags that don't get loaded right away

react-example/src/index.tsx
```diff
@@ -43,7 +43,7 @@ const HomeLink = styled(Link)`
     ({ email }) => {
       return axios
         .post(
-          'http://localhost:5000/generate',
+          process.env.JWT_GENERATOR || '',
```
mprew97, line 46: would it be best to at least leave the default as `http://localhost:5000/generate` so the api hits something?

mprew97 (approved):
Love all of this, Paul! Thanks for the refactor here. Left one comment that's a nit, but otherwise lgtm.

PR #308, opened 2024-02-26
MOB 7586 Diff/Merge Messages

hani-iterable:
## JIRA Ticket(s) if any

* [MOB-7586](https://iterable.atlassian.net/browse/MOB-7586)

> Additionally, we resolved test case failures resulting from these changes.

[MOB-7586]: https://iterable.atlassian.net/browse/MOB-7586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mprew97:
I'm thinking that we may want to allow customers to pass in their `currentMessageIds` here as opposed to us keeping a list in local storage. I can see a common case where the messages a user has stored somewhere is different or not up to date with what we have in `localStorage`. This puts the onus on them to handle any state changes to their messages.

@bradumbaugh any thoughts here?

bradumbaugh:
@mprew97 Hm, I think there may be cases where the developer isn't even aware of all the messages that have been fetched? Like maybe a few messages come back, but for a placement that hasn't ever been loaded, so the client codenever interacts with those messages... it's pretty convenient to have this happen behind the scenes.

bradumbaugh:
Also, we should probably caution against storing embedded messages in a separate location, to prevenet the display of messages for which the user is no longer eligible.

mprew97:
@hardikmashru can you resolve the conflicts on this when you have a chance? it's ready for review/merge otherwise.

hardikmashru:
> @hardikmashru can you resolve the conflicts on this when you have a chance? it's ready for review/merge otherwise.

Done

hardikmashru:
Hi @mprew97 it is resolved now.

PR #315, opened 2024-03-01
MOB-7859 (Long buttons become stacked)

hani-iterable:
## JIRA Ticket(s) if any

* [MOB-7859](https://iterable.atlassian.net/browse/MOB-7859)

## Description

> Here we have set the button width according to the requirement.

### DIFF
--- a/src/components/banner/index.tsx
+++ b/src/components/banner/index.tsx
@@ -68,10 +68,14 @@ export const Banner = (props: IBannerProps) => {
     fontSize: '16px',
     marginBottom: '16px'
   };
-  const bannerButtons = {
-    marginTop: '20px'
+  const bannerButtons: CSSProperties = {
+    marginTop: '20px',
+    flex: '1',
+    flexDirection: 'row',
+    display: 'flex'
   };
   const defaultButtonStyles: CSSProperties = {
+    maxWidth: 'calc(50% - 32px)',
     textAlign: 'left',
     fontSize: '16px',
     fontWeight: 'bold',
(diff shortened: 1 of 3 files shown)


src/components/banner/index.tsx
```diff
@@ -68,18 +68,24 @@ export const Banner = (props: IBannerProps) => {
...
-  const bannerButtons = {
-    marginTop: '20px'
+  const bannerButtons: CSSProperties = {
+    marginTop: '20px',
+    flex: '1',
+    flexDirection: 'row',
+    display: 'flex'
   };
   const defaultButtonStyles: CSSProperties = {
+    maxWidth: 'calc(50% - 32px)',
```
mprew97, line 78: why 32px? would prefer some percentage here instead so it's more responsive unless this needs to be a fixed pixel value.

src/components/card/index.tsx
```diff
@@ -70,6 +70,7 @@ export const Card = (props: ICardProps) => {
     marginBottom: '16px'
   };
   const defaultButtonStyles: CSSProperties = {
+    maxWidth: 'calc(50% - 32px)',
```
mprew97, line 73: same here

src/components/notification/index.tsx
```diff
@@ -46,23 +46,28 @@ export const Notification: React.FC<NotificationProps> = ({
   };
 
   const primaryButtonDefaultStyle: CSSProperties = {
+    maxWidth: 'calc(50% - 32px)',
```
mprew97, line 49: and here

mprew97 (changes requested):

src/components/banner/index.tsx
hani-iterable: We've added a margin for the button, so using a percentage might not function correctly in this context.

src/components/card/index.tsx
hani-iterable: We've added a margin for the button, so using a percentage might not function correctly in this context.

src/components/notification/index.tsx
hani-iterable: We've added a margin for the button, so using a percentage might not function correctly in this context.

mprew97 (approved):

PR #328, opened 2024-03-18
Mob 8058 (Web SDK: Click Handling)

hani-iterable:
## JIRA Ticket(s) if any

* [MOB-8058](https://iterable.atlassian.net/browse/MOB-8058)

## Description

> We have implemented Web click handling for embedded OOTB views.

[MOB-8058]: https://iterable.atlassian.net/browse/MOB-8058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/embedded/embeddedManager.ts
```diff
@@ -141,4 +154,62 @@ export class EmbeddedManager {
...
+    const payload = {
+      messageId: message?.metadata?.messageId,
+      campaignId: message?.metadata?.campaignId
+    };
+
+    const emailOrUserId =
+      (localStorage.getItem(SHARED_PREF_EMAIL) as string) ??
+      (localStorage.getItem(SHARED_PREF_USER_ID) as string);
+
+    console.log('email', emailOrUserId);
```
mprew97, line 205: can we remove the console statement?

src/embedded/embeddedManager.ts
```diff
@@ -141,4 +154,62 @@ export class EmbeddedManager {
...
+      IterableActionRunner.executeAction(
+        null,
+        iterableAction,
+        IterableActionSource.EMBEDDED
+      );
+    }
+  }
+
+  trackEmbeddedClick(
+    message: any,
```
mprew97, line 192: shouldn't this be of type `EmbeddedMessage`?

src/embedded/embeddedManager.ts
```diff
@@ -141,4 +154,62 @@ export class EmbeddedManager {
...
+    buttonIdentifier: string | '',
+    clickedUrl: string | ''
```
mprew97, line 194: we don't need the ` | ''` since that is just a string.

src/embedded/embeddedManager.ts
```diff
@@ -141,4 +154,62 @@ export class EmbeddedManager {
...
+  ) {
+    const payload = {
+      messageId: message?.metadata?.messageId,
+      campaignId: message?.metadata?.campaignId
+    };
+
+    const emailOrUserId =
```
mprew97, line 201: We shouldn't be tracking the click if there is nothing found in local storage. Shouldn't happen, but we should account for it. And ideally, `email` and `userId` should be fields passed in to this method so that the search for local storage is handled outside the scope of this method.

mprew97 (changes requested):
Thanks a ton for the quick turnaround here! Have a few comments to address.

react-example/src/views/EmbeddedMsgs.tsx
```diff
@@ -141,98 +124,36 @@ export const EmbeddedMsgs: FC<Props> = () => {
...
-                      handleOpenUrl(
-                        secondaryButton?.action?.type,
-                        secondaryButton?.action?.data
-                      )
-                    }
-                    onClickView={() =>
-                      handleOpenUrl(defaultAction?.type, defaultAction?.data)
-                    }
+                    parentStyle={{ margin: 0 }}
+                    messageData={data}
```
bradumbaugh, line 135: Could we clarify this? Maybe call the Card's attribute `message` (instead of `messageData`) and then pass in a `message`, rather than `data`? The idea here is that we're passing in the message to be rendered... I think talking about "data" removes a bit of clarity

react-example/src/views/EmbeddedMsgs.tsx
```diff
@@ -141,98 +124,36 @@ export const EmbeddedMsgs: FC<Props> = () => {
...
 
               case 1:
                 return (
                   <Banner
                     key={index.toString()}
-                    BannerStyle={{ margin: 0 }}
-                    title={data?.title}
-                    text={data?.body}
-                    imgSrc={data?.mediaUrl}
+                    parentStyle={{ margin: 0 }}
```
bradumbaugh, line 143: What is "parentStyle"? Just want to be sure I understand what this is / why it's named that way

react-example/src/views/EmbeddedMsgs.tsx
```diff
@@ -141,98 +124,36 @@ export const EmbeddedMsgs: FC<Props> = () => {
...
+                    messageData={data}
```
bradumbaugh, line 144: Same comment as above about `messageData`. Won't keep calling this out, but I think my comment is valid for banner / notification / card

react-example/src/views/EmbeddedMsgs.tsx
```diff
@@ -141,98 +124,36 @@ export const EmbeddedMsgs: FC<Props> = () => {
...
-                      handleOpenUrl(
-                        secondaryButton?.action?.type,
-                        secondaryButton?.action?.data
-                      )
-                    }
-                    onClickView={() =>
-                      handleOpenUrl(defaultAction?.type, defaultAction?.data)
-                    }
-                  />
+                  <Notification key={index.toString()} messageData={data} />
```
bradumbaugh, line 156: messageData vs message again

src/components/card/index.tsx
```diff
@@ -1,48 +1,21 @@
...
-  secondaryBtnStyle?: CSSProperties;
-  secondaryDisableBtnStyle?: CSSProperties;
-  onClickView?: () => void;
-}
+import { EmbeddedManager, TextParentStyles } from 'src/index';
+import { EmbeddedMessageData } from '../types';
 
 /* Note: Add export to this const when support Embedded Message View Types in a later release. */
-const Card = (props: ICardProps) => {
+export const Card = (props: EmbeddedMessageData) => {
```
bradumbaugh, line 6: `IterableEmbeddedMessage` instead of `EmbeddedMessagData`?

src/components/notification/index.tsx
```diff
@@ -1,42 +1,21 @@
...
-  onClickSecondaryBtn,
-  textStyle,
-  titleStyle,
-  primaryDisableBtnStyle,
-  secondaryDisableBtnStyle,
-  disablePrimaryBtn,
-  disableSecondaryBtn,
-  onClickView
-}) => {
+export const Notification = (props: EmbeddedMessageData) => {
```
bradumbaugh, line 6: `IterableEmbeddedMessage` instead of `EmbeddedMessageData`?

src/utils/IterableActionRunner.ts
```diff
@@ -0,0 +1,62 @@
...
+        );
+      }
+    }
+    return false;
+  }
+}
+
+export class IterableActionRunner extends IterableActionRunnerImpl {
+  static config = IterableConfig;
+}
```
bradumbaugh, line 62: @hani-iterable What is this needed for?

bradumbaugh, line 62: @hani-iterable What is this config needed for?

bradumbaugh (changes requested):
Left some minor comments / questions

mprew97:
Also, this is kind of a late catch, but can we remove instances of manually setting email and user id on local storage in the Web SDK source? We already have methods that customers/devs use to do that. You can see we have a `setEmail` and `setUserId` on the `initialize()` method we are already using for embedded in the react-example app. 

We should just be able to fetch from local storage without having to set it, except for when `setUser` and `setUserId` is called. 

See the method definition in `authorization.ts`. To use it, you would just call

```const { setEmail, setUserId } = initialize(...)```

File: src/embedded/embeddedManager.ts (hunk 1 above)
Comment (hani-iterable) [reply src/embedded/embeddedManager.ts:210]: Done

File: src/embedded/embeddedManager.ts (hunk 2 above)
Comment (hani-iterable) [reply src/embedded/embeddedManager.ts:197]: I have set 'IterableEmbeddedMessage' as it already exists and match with the Android SDK.

File: src/embedded/embeddedManager.ts (hunk 3 above)
Comment (hani-iterable) [reply src/embedded/embeddedManager.ts:199]: Done

File: src/embedded/embeddedManager.ts (hunk 4 above)
Comment (hani-iterable) [reply src/embedded/embeddedManager.ts:206]: I've implemented a check for both the `email` and `userId` in the Android SDK, following the same approach used in other methods. During API calls, the `email` and `userId` are added to the request, ensuring consistency across our SDK architectures.

File: react-example/src/views/EmbeddedMsgs.tsx (hunk 5 above)
Comment (hani-iterable) [reply react-example/src/views/EmbeddedMsgs.tsx:135]: Done

File: react-example/src/views/EmbeddedMsgs.tsx (hunk 6 above)
Comment (hani-iterable) [reply react-example/src/views/EmbeddedMsgs.tsx:143]: The 'parentStyle' will set the style to the parent view of all the components like background,  borderRadius, etc.

File: react-example/src/views/EmbeddedMsgs.tsx (hunk 7 above)
Comment (hani-iterable) [reply react-example/src/views/EmbeddedMsgs.tsx:144]: Done

File: src/components/banner/index.tsx [hunk 12]
```diff
@@ -1,47 +1,21 @@
...
-  secondaryBtnStyle?: CSSProperties;
-  secondaryDisableBtnStyle?: CSSProperties;
-  onClickView?: () => void;
-}
+import { EmbeddedManager } from 'src/index';
+import { EmbeddedMessageData } from '../types';
 
 /* Note: Add export to this const when support Embedded Message View Types in a later release. */
-const Banner = (props: IBannerProps) => {
+export const Banner = (props: EmbeddedMessageData) => {
```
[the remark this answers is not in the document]
hani-iterable: We can't. `IterableEmbeddedMessage` is responsible for determining the message type received, while `EmbeddedMessageData` contains the properties needed for custom styling of Out-of-the-Box (OOTB) views.

src/components/card/index.tsx
hani-iterable: We can't. `IterableEmbeddedMessage` is responsible for determining the message type received, while `EmbeddedMessageData` contains the properties needed for custom styling of Out-of-the-Box (OOTB) views.

src/components/notification/index.tsx
hani-iterable: We can't. `IterableEmbeddedMessage` is responsible for determining the message type received, while `EmbeddedMessageData` contains the properties needed for custom styling of Out-of-the-Box (OOTB) views.

src/utils/IterableActionRunner.ts
hani-iterable: No need for that. I have removed.

hani-iterable:
> initialize()

Got it, you're asking about the use of the 'initialize()' method. Do we need to pass the userID from the Sample app when calling any method related to Embedded messages, or is there another way to handle this?

mprew97:
> > initialize()
> 
> @mprew97 Got it, you're asking about the use of the 'initialize()' method. Do we need to pass the userID from the Sample app when calling any method related to Embedded messages, or is there another way to handle this?

@hani-iterable For API requests, once you call setEmail or setUserId, those automatically get tacked on to all requests. In authorization.ts, see lines 237 - 328. We intercept any request using baseAxiosRequest and add the email or userId to the req.

PR #330, opened 2024-03-19
Fix embedded events api calls

mprew97:
## JIRA Ticket(s) if any
api calls were not being intercepted and injected with the proper email or userId

## Test Steps


jyu115 (approved):
lgtm

PR #354, opened 2024-04-16
OOTB Rework: Need to be pure JS/HTML/CSS

hardikmashru:
## JIRA Ticket(s) if any

## Test Steps


src/components/banner/index.tsx
```diff
@@ -82,93 +84,127 @@ export const Banner = (props: EmbeddedMessageData) => {
...
+    if (bannerDiv) {
+      bannerDiv.addEventListener('click', handleBannerClick);
+    }
+    if (primaryButtonClick) {
+      addButtonClickEvent(primaryButtonClick, 0);
+    }
+    if (secondaryButtonClick) {
+      addButtonClickEvent(secondaryButtonClick, 1);
+    }
+  }, 0);
```
mprew97, line 137: just curious why this is a timeout?

src/components/banner/index.tsx
```diff
@@ -82,93 +84,127 @@ export const Banner = (props: EmbeddedMessageData) => {
...
+            style="${defaultImageStyles}; ${imgStyle || ''}" src="${
+                message?.elements?.mediaUrl
+              }" />`
+            : ''
+        }
+      </div>
+      <div class="banner" 
+       style="${bannerButtons}">
+        ${message?.elements?.buttons
+          ?.map((button: any, index: number) => {
```
mprew97, line 174: We should know the type of the button. assume we have it in the code somewhere. if not, let me know and can find that definition for you so you can type it.

src/components/banner/index.tsx
```diff
@@ -82,93 +84,127 @@ export const Banner = (props: EmbeddedMessageData) => {
...
+            const buttonStyle =
+              index === 0 ? primaryBtnStyle : secondaryBtnStyle;
```
mprew97, line 176: we have a lot of logic for the styles based on primary/secondary button. 

we should create an object that handles this outside of the component string to clean it up a bit. It would look sort of like:

```
const getObj = () => ({
   buttonStyle: index === 0 ? primaryBtnStyle : secondaryBtnStyle,
   disablePrimaryBtn: index === 0 ? 'disabled' : 'enabled',
   ...
})
```

src/components/card/index.tsx
```diff
@@ -1,166 +1,218 @@
...
+          ?.map((button: any, index: number) => {
+            const buttonStyle =
+              index === 0 ? primaryBtnStyle : secondaryBtnStyle;
+            const disableStyle =
+              index === 0 ? primaryDisableBtnStyle : secondaryDisableBtnStyle;
+            return `
+              <button 
+                key="${index}" 
+                ${
+                  index === 0
```
mprew97, line 182: same comment about the button styles as above

mprew97 (changes requested):
A few questions/comments here, but mostly looks great! 

Since we are no longer using react for these components, lets remove that from our package.json file and the yarn.lock file so it's not included in the GA release.

src/components/banner/index.tsx
```diff
@@ -82,93 +84,127 @@ export const Banner = (props: EmbeddedMessageData) => {
...
+    <style>${mediaStyle}</style>
+    <div 
+      class="banner" 
+      id="${parentId}"
+      name="${message?.metadata?.messageId}-banner"
+      style="${defaultBannerStyles}; ${parentStyle || ''}" 
+    >
+      <div class="banner" 
+      style="display: flex; flex-direction: row;">
+        <div class="banner"
```
mprew97, line 149: can we provide an `id` to these two divs?

mprew97 (commented):
one more thought, all `divs` and other elements should have an `id` tag for css targeting.

src/components/banner/index.tsx
hardikmashru: There is a concept in js called Event loop. It manages asynchronous operations. The event loop works by constantly monitoring two queues: the call stack and the callback queue. When the event loop starts, it checks the call stack. If the call stack is empty, the event loop will take the next function from the callback queue and move it to the call stack. The function will then be executed.

In this code, setTimeout schedules the execution of the block after the current stack clears, allowing time for DOM elements to render. Without setTimeout, the code executes synchronously, attempting to access DOM elements before they exist, leading to undefined results.
You can learn about event loop with example here: https://www.jsv9000.app/

src/components/banner/index.tsx
hardikmashru: Done

src/components/banner/index.tsx
hardikmashru: Done

src/components/card/index.tsx
hardikmashru: Done

src/components/banner/index.tsx
hardikmashru: Done

src/components/banner/index.tsx
mprew97: Got it, it's just there to handle the case where the dom isn't rendered yet. In that case, we should allow our customer's to handle the hook up of the event handling. 

Meaning that the sample app should do this attachment but it shouldn't be done in this component. 

cc: @bradumbaugh so you are aware and can provide input as necessary.

hardikmashru: Hi @mprew97 I am NOT sure what you mean but we need to handle these click events inside the components because inside click events of the buttons and the main view we handle clicked url and also the tracking of embedded click.

mprew97: hmmm, i don't think `setTimeout` is necessary still. You should theoretically just be able to create a function that gets called after the component is rendered. `setTimeout` seems like overkill https://teamtreehouse.com/community/how-do-i-set-an-event-listener-on-an-element-that-doesnt-yet-exist

mprew97:
Once we address the `setTimeout` discussion, this one is good to merge, @hardikmashru.

src/components/banner/index.tsx
hardikmashru: Hi @mprew97 We have been trying alternative approaches, but unfortunately, we are still facing multiple click issues with other approaches. Regarding the setTimeout function, we won't have any network speed-related problems as we are rendering components based on message data that has no dependency on the network. Indeed, setTimeout is not the ideal solution, so we will try for a more appropriate one.

Issue #356, opened 2024-04-22
How to connect to European Data Center with the web SDK?


mistyharsh:
I am having difficulty in connecting to the Iterable's European data center. I tried using NPM approach as well as CDN approach as documented [here][1]. But, SDK always connects to the default endpoint (`https://api.iterable.com/api`).

As the guidelines says:

> To do this, on the web server that hosts your site, set the IS_EU_ITERABLE_SERVICE environment variable to true.

I used following different means to setup this flag.
- `process.env.IS_EU_ITERABLE_SERVICE = true`.
- `globalThis.IS_EU_ITERABLE_SERVICE = true`.
- `window.IS_EU_ITERABLE_SERVICE = true`.

### CDN approach.

I used following script to initialize the Iterable SDK:

```
<script src="https://unpkg.com/@iterable/web-sdk/index.js"></script>
```

But as I see in this compiled code, the flag is used as shown below:

```
var t="https://".concat("true"==={VERSION:"1.1.0-beta"}.IS_EU_ITERABLE_SERVICE?"api.eu.iterable.com":"api.iterable.com","/api")
```

As far as I understand, the condition will always be false and would always pick the default `api.iterable.com` URL,

### NPM approach

With NPM approach, I get the same requests. I tried using both files specified by `main` and `browser` field in the package.json file of the package:
- package.json `main` field points to `index.node.js` file.
- package.json `browser` field points to `index.js` file.

We are using two bundlers - Vite and Webpack. And, we can configure to to pick any file. But it looks like the code is already compiled where the `IS_EU_ITERABLE_SERVICE` flag is never initialized anywhere in the compiled code. Both files contain the code similar to the one served by CDN.

---

I used versions `1.0.8` and latest `1.1.0-beta`. Is there any other solution for this?

[1]: https://github.com/Iterable/iterable-web-sdk?tab=readme-ov-file#iterables-european-data-center-eudc

mprew97:
Hi @mistyharsh, currently investigating this. Thanks for the writeup.

mprew97:
@mistyharsh we are cutting a new release. once that is out, please use that version and try again and your issue should be resolved :)

mistyharsh:
Hey @mprew97,

I tried this last week and the issue still persists with version `v.1.0.9` if I follow official documentation. I tested this with Vite and defined the environment variable like this:

```ts
import { defineConfig } from 'vite';

export default defineConfig({
  define: {
[...]

PR #360, opened 2024-04-25
Pass in additional env vars to webpack for EU

mprew97:
## Description
Will solve https://github.com/Iterable/iterable-web-sdk/issues/356

## Test Steps

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,7 +7,8 @@ function getParsedEnv() {
   if (!env.error) {
     return {
       ...env.parsed,
-      VERSION: version
+      VERSION: version,
+      IS_EU_ITERABLE_SERVICE: process.env.IS_EU_ITERABLE_SERVICE || false,
     };
   }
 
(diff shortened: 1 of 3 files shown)


ts-nguyen (approved):

martinmckenna:
just a note here (because I still get notifications for this project 😃), environment variables were never meant to be exposed or able to be set by the consumer of the SDK. In fact, if the user is using vite instead of webpack, `process.env` isn't even a thing that exists. It's called `import.meta.env`: https://vitejs.dev/guide/env-and-mode#env-variables

And if they're using a different bundler or no bundler at all, it's even less clear how you update that value.

Updating basic config values was what the config setter/getter was meant to do: https://github.com/Iterable/iterable-web-sdk/blob/main/src/utils/config.ts. Setting the EU datacenter should really be done via this. Environment variables were only meant to be private values set when running/building the SDK

mistyharsh:
@martinmckenna Thanks for the explanation. That's the apt behavior. Libraries, especially, frontend libraries should not assume `process.env` or `import.meta.env`. Those should be accepted as parameters via some init/config functions or at the least use `globalThis`.

PR #362, opened 2024-04-26
Rework Embedded Session Manager

hardikmashru:
## JIRA Ticket(s) if any

* [MOB-8136](https://iterable.atlassian.net/browse/MOB-8136)

## Test Steps

[MOB-8136]: https://iterable.atlassian.net/browse/MOB-8136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mprew97:
@bradumbaugh for your vis

mprew97:
One thought I have here, specifically after working in the `embeddedSessionManager` some more, could we just add a field on to the `EmbeddedSession` class `impressions` where we track and store impressions? An impression has to exist as part of a session so it makes sense to make it a part of that class and not have two separate objects to keep track of.

hardikmashru:
Hi @mprew97 yes that is what I thought so but I guess we had followed the same code as Android when we worked on web implementation. Let me know if you need me to work on your suggestion before you merge this.

PR #365, opened 2024-04-30
[MOB-8148]:  clean up embedded types

mprew97:
## JIRA Ticket(s) if any

* [MOB-8148](https://iterable.atlassian.net/browse/v)

## Description

- cleans up types and track event request logic
- removes react dep from main SDK (not necessary)
- removes package lock file (only need yarn). 

## Test Steps

[MOB-8148]: https://iterable.atlassian.net/browse/MOB-8148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/embedded/embeddedSessionManager.ts
```diff
@@ -75,35 +77,38 @@ export class EmbeddedSessionManager {
     }
 
     if (this.impressions.size) {
+      const sessionPayload: EmbeddedSessionRequestPayload = {
+        session: {
+          start: this.session.start?.getTime(),
+          end: new Date().getTime(),
+          id: this.session.id
+        },
+        impressions: this.session.impressions,
```
hardikmashru, line 86: @mprew97 I think here we use this.impressions as we set the data in instance variable instead of session one in line: 104

src/events/embedded/types.ts
```diff
@@ -1,116 +1,81 @@
...
+interface ButtonAction {
   type: string;
   data?: string;
 }
-export interface IEmbeddedMessageElements {
+
+interface Button {
+  id: string;
+  title: string;
+  action?: ButtonAction;
```
pauljung14, line 21: Is the action prop ever expected to not exist a button? If a button is defined, wouldn't it have either an openUrl or customAction action?

src/events/embedded/types.ts
```diff
@@ -1,116 +1,81 @@
...
   title?: string;
   body?: string;
   mediaUrl?: string;
 
-  buttons?: [IEmbeddedMessageElementsButton];
-  text?: [IEmbeddedMessageElementsText];
-  defaultAction?: IEmbeddedMessageElementsDefaultAction;
+  buttons?: Button[];
+  text?: Text[];
+  defaultAction?: DefaultAction;
```
pauljung14, line 31: Similar comment here. Even when set to "none" there's a value we pass in for that, right?

src/events/embedded/types.ts
```diff
@@ -1,116 +1,81 @@
...
-  metadata: IEmbeddedMessageMetadata;
-  elements?: IEmbeddedMessageElements;
-  payload?: Array<any>;
-}
-
-export interface IEmbeddedMessage {
+export interface EmbeddedMessage {
   email?: string;
   userId?: string;
   messageId: string;
```
pauljung14, line 54: Looks redundant to me that we have messageId here and within metadata... not sure if that's the case for our mobile sdks too though...

src/embedded/embeddedSessionManager.ts
mprew97: @hardikmashru do you mean `this.getImpressionList` should be used instead? Not sure if i follow you.

src/events/embedded/types.ts
```diff
@@ -1,116 +1,81 @@
...
-  metadata?: IEmbeddedMessageMetadata;
-  elements?: IEmbeddedMessageElements;
+  metadata: Metadata;
+  elements?: Elements;
   payload?: Array<any>;
```
pauljung14, line 57: Is Array<any> the best type to use here? I think a tighter type for embedded payload would be Record<string, any>

src/embedded/embeddedManager.ts
```diff
@@ -114,9 +111,9 @@ export class EmbeddedManager {
       this.notifyUpdateDelegates();
     }
     for (let i = 0; i < msgsList.length; i++) {
-      const messages = {} as IEmbeddedMessageData;
-      messages.messageId = msgsList[i].metadata.messageId;
-      await trackEmbeddedMessageReceived(messages);
+      const message = {} as EmbeddedMessage;
+      message.messageId = msgsList[i].metadata.messageId;
```
pauljung14, line 115: Is message.messageId always set like this? Across our mobile sdks too?

src/embedded/embeddedManager.ts
```diff
@@ -114,9 +111,9 @@ export class EmbeddedManager {
...
+      await trackEmbeddedReceived(message.messageId);
```
pauljung14, line 116: Also, it doesn't look like we're using message anywhere further (though I'm not in an IDE so can't say for certain), and the typecasting doesn't look like it's doing anything (it's not like message.messageId is a special type, just a string), so perhaps could be simplified to:
```suggestion
      const { messageId } = msgsList[i].metadata;
      await trackEmbeddedReceived(messageId);
```

src/embedded/embeddedSessionManager.ts
mprew97: gotcha

src/events/embedded/events.ts
```diff
@@ -1,63 +1,102 @@
...
+  baseIterableRequest<IterableResponse>({
     method: 'POST',
     url: EndPoints.msg_received_event_track,
-    data: payload,
+    data: {
+      messageId,
+      deviceInfo: {
+        platform: WEB_PLATFORM,
+        deviceId: global.navigator.userAgent || '',
+        appPackageName: appPackageName || window?.location?.hostname
```
pauljung14, line 29: Curious, what is the hostname referring to that it should be assigned to appPackageName?

src/events/embedded/types.ts
mprew97: it's optional so it is possible

src/events/embedded/types.ts
mprew97: good catch

pauljung14 (approved):

src/embedded/embeddedClickEvents.ts
```diff
@@ -1,30 +1,33 @@
...
-  EmbeddedMessageElementsButton,
+  IterableEmbeddedManager,
+  IterableEmbeddedButton,
   IterableEmbeddedMessage
 } from '../embedded';
 
-const embeddedManager = new EmbeddedManager();
-export const handleElementClick = (message: IterableEmbeddedMessage) => {
+export const handleElementClick = (
+  embeddedManager: IterableEmbeddedManager,
```
bradumbaugh, line 8: @mprew97 Should we be exporting the things in this file? I think these should only be used under the hood by the OOTB classes, not by customers...

mprew97: they need to be exported by this file. We can lock them down from being exported outside of the SDK as a whole though, if that makes sense. Great callout, will go ahead and do that.

src/embedded/embeddedSessionManager.ts
```diff
@@ -1,52 +1,48 @@
...
-  public session: IEmbeddedSession = new EmbeddedSession(
-    undefined,
-    undefined,
-    '0',
-    undefined
-  );
+export class IterableEmbeddedSessionManager {
+  public appPackageName: string;
+  private impressions: Map<string, EmbeddedImpression> = new Map();
+  public session: EmbeddedSession = new EmbeddedSession();
```
bradumbaugh, line 41: @mprew97 Something that trips me up here — it feels to me like the impressions should be part of the session, since that's how we track them in the API

bradumbaugh: Either that, or... why do we need a session object? If a session's impressions are being tracked at the top level of the session manager... should everything else about a session be tracked that way, too?

src/embedded/types.ts
```diff
@@ -1,12 +1,53 @@
-export interface EmbeddedMessageUpdateHandler {
+export interface IterableEmbeddedButtonAction {
+  type: string;
+  data?: string;
+}
+
+interface IterableEmbeddedElements {
```
bradumbaugh, line 6: @mprew97 Did we decide to export all these?

src/embedded/embeddedSessionManager.ts
mprew97: @hardikmashru had replied on a separate PR that they were following the Android convention. Happy for us to swap over if we think it's necessary. 

and tbf, the request body of `embedded/session` has `impressions` and `session` as siblings for the parameters... likely why they are separated.

bradumbaugh: @mprew97 OK, spent some time looking at the API and the code, and I follow what you're saying. Thanks!

bradumbaugh (approved):
Left a few minor comments, but this looks like a good direction to me. Thanks.

src/embedded/embeddedClickEvents.ts
mprew97: Just verified, these aren't being exported by the SDK as it stands.

src/embedded/types.ts
mprew97: yep, miss on my part. done now.

PR #374, opened 2024-05-07
easier UUID testing in sample app

mprew97:
## JIRA Ticket(s) if any

* [MOB-8426](https://iterable.atlassian.net/browse/MOB-8426)

## Description

Makes it a bit easier to test UserID behavior in the React example app by allowing you to set which user type to use on login (email or userId).

## Test Steps
1. Regression test that email still works with an email based project
2. Update environment variables to point to a UUID based project
3. Login and make api calls with UUID as opposed to email

### DIFF
--- a/react-example/src/components/LoginForm.tsx
+++ b/react-example/src/components/LoginForm.tsx
@@ -1,4 +1,4 @@
-import { FC, FormEvent, useState } from 'react';
+import { ChangeEvent, FC, FormEvent, useState } from 'react';
 import styled from 'styled-components';
 
 import _TextField from 'src/components/TextField';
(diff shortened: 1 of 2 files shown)


react-example/src/components/LoginForm.tsx
```diff
@@ -42,12 +55,14 @@ export const LoginForm: FC<Props> = ({ setEmail, logout, refreshJwt }) => {
...
+    const setUser = useEmail ? setEmail : setUserId;
+
+    setUser(user)
       .then(() => {
         setEditingUser(false);
-        setLoggedInUser({ type: 'user_update', data: email });
+        setLoggedInUser({ type: 'user_update', data: user });
       })
-      .catch(() => updateEmail('Something went wrong!'));
+      .catch(() => updateUser('Something went wrong!'));
```
pauljung14, line 65: Looking at this, it's kinda strange that we update the user value to this error message. Can consider refactoring at some point to show an inline error below the field or open an alert or something. Just a thought

react-example/src/components/LoginForm.tsx
```diff
@@ -56,60 +71,100 @@ export const LoginForm: FC<Props> = ({ setEmail, logout, refreshJwt }) => {
...
   };
 
+  const handleRadioChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setUseEmail(e.target.value === 'email');
+  };
+
   const first5 = loggedInUser.substring(0, 5);
   const last9 = loggedInUser.substring(loggedInUser.length - 9);
 
+  const editingUserForm = (
```
pauljung14, line 94: Since these are basically mini-components, could we use React component formatting for added clarity? 
```suggestion
  const EditingUserForm = () => (
```
and then update it in line 157 to `<EditingUserForm />`

pauljung14: Similarly for `userIdOrEmailForm` and other components below

pauljung14 (approved):
Some formatting nits

react-example/src/components/LoginForm.tsx
mprew97: yeah, great idea

PR #376, opened 2024-05-09
[MOB-8425]: strengthen network interceptors

mprew97:
## JIRA Ticket(s) if any

* [MOB-8425](https://iterable.atlassian.net/browse/MOB-8425)

## Description

Our network interceptors are prone to misses whenever new routes are added.

This PR makes it so that all places in the SDK calling our API are pulling from a consistent route object and that same route object is being used by the interceptors. This way, any new routes added to there will have to be made sure to work with our interceptors. 

## Test Steps
Start sample app
Test all API calls and make sure userID and email are properly attached. 

[MOB-8425]: https://iterable.atlassian.net/browse/MOB-8425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/constants.ts
```diff
@@ -62,6 +59,120 @@ export const URL_SCHEME_OPEN = 'openUrl';
...
+export type RouteConfig = {
+  route: string;
+  body: boolean; // true for POST/PUT requests
+  current: boolean; // true if email or userId in request needs to be prepended with `current`
+  nestedUser: boolean; // true if route expects email or userId field to be nested in user object
+};
+
+type EndPointStructure = Record<string, RouteConfig>;
+
+export const END_POINTS: EndPointStructure = {
```
pauljung14, line 71: Isn't endpoint one word?

src/constants.ts
```diff
@@ -62,6 +59,120 @@ export const URL_SCHEME_OPEN = 'openUrl';
 export const SHARED_PREF_USER_ID = 'userId';
 export const SHARED_PREF_EMAIL = 'email';
 
+export type RouteConfig = {
+  route: string;
+  body: boolean; // true for POST/PUT requests
+  current: boolean; // true if email or userId in request needs to be prepended with `current`
+  nestedUser: boolean; // true if route expects email or userId field to be nested in user object
+};
```
pauljung14, line 67: Rec to use jsdoc format so that hovering over properties can show these comments elsewhere
```suggestion
export type RouteConfig = {
  route: string;
  /** true for POST/PUT requests */
  body: boolean;
  /** true if email or userId in request needs to be prepended with `current` */
  current: boolean;
  /** true if route expects email or userId field to be nested in user object */
  nestedUser: boolean;
};
```

src/constants.ts
```diff
@@ -20,9 +20,6 @@ export const BASE_URL = process.env.BASE_URL || ITERABLE_API_URL;
 
 export const GETMESSAGES_PATH = '/inApp/web/getMessages';
 
-/** @todo update once new endpoint is ready */
-export const CACHE_ENABLED_GETMESSAGES_PATH = '/newEndpoint';
```
pauljung14, line 24: In `/inapp/request.ts`, line 25, would you mind updating the commented out line to read "tbd" or something to replace the reference to the placeholder constant?

pauljung14 (approved):

src/constants.ts
mprew97: it is haha, good callout. will update

PR #377, opened 2024-05-10
fix focus bug and console error

mprew97:
## JIRA Ticket(s) if any

* [MOB-8504](https://iterable.atlassian.net/browse/MOB-8504)
* [MOB-8469](https://iterable.atlassian.net/browse/MOB-8469)

## Description
Fix sample app console error
Fix login field focus issue

## Test Steps


react-example/src/components/LoginForm.tsx
pauljung14, line 1: If you need to pass in `isEditingUser` to these subcomponents, you could set up props to pass in for them like you would any component, but putting it back in the main jsx return is fine too. Just thought since you already split them out, not too big of a change to add props to them, up to you

pauljung14 (approved):

PR #380, opened 2024-05-13
[MOB-8515]: allow-popups-to-escape-sandbox

mprew97:
## JIRA Ticket(s) if any

* [MOB-8515](https://iterable.atlassian.net/browse/MOB-8515)

## Description

See https://googlechrome.github.io/samples/allow-popups-to-escape-sandbox/

## Test Steps

[MOB-8515]: https://iterable.atlassian.net/browse/MOB-8515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/inapp/utils.ts
```diff
@@ -284,9 +284,14 @@ const generateSecuredIFrame = () => {
   iframe.setAttribute('id', 'iterable-iframe');
   // allow-popups and allow-top-navigation is to enable links for Safari since the iframe will block
   // event handlers on elements in it preventing our custom link handling
+  console.log({ p: process.env });
```
ts-nguyen, line 287: Should remove console log

ts-nguyen (commented):
Do we need to update the README for this change and make a disclaimer for this config. I imagine we'll need a security review on this and possibly legal?

src/inapp/utils.ts
mprew97: ope, good catch.

mprew97:
> Do we need to update the README for this change and make a disclaimer for this config. I imagine we'll need a security review on this and possibly legal?

@ts-nguyen Security is aware of this and will review this today. As for the README, I've alerted docs about this PR so will let them determine what changes are needed.

webpack.config.js
```diff
@@ -9,6 +9,8 @@ function getParsedEnv() {
       ...env.parsed,
       VERSION: version,
       IS_EU_ITERABLE_SERVICE: process.env.IS_EU_ITERABLE_SERVICE || false,
+      DANGEROUSLY_ALLOW_JS_POPUP_EXECUTION:
+        process.env.DANGEROUSLY_ALLOW_JS_POPUP_EXECUTION || false
```
ts-nguyen, line 13: Should we do a strict check against `true` given the effect of the changes this value controls?

ts-nguyen (approved):
Left an additional suggestion for a safer check!

webpack.config.js
mprew97: this is the webpack config to allow the env variable to be passed in. made the check a const and made it stricter [here](https://github.com/Iterable/iterable-web-sdk/pull/380/files#diff-8fa4b52909f895e8cda060d2035234e0a42ca2c7d3f8f8de1b35a056537bf199R10)

react-example/.env.example
```diff
@@ -8,3 +8,6 @@
 
 # Convenience variable to automatically set the login email during testing.
 # LOGIN_EMAIL=email@email.com
+
+# IS_EU_ITERABLE_SERVICE=false
+# DANGEROUSLY_ALLOW_JS_POPUP_EXECUTION=false
```
pauljung14, line 13: Why do we need this at the sample app level?

jberry-iterable (approved):
Thank you for making this a config change!

react-example/.env.example
mprew97: When the package is installed via package manager, devs don't have access to modify the `.env` file of the web-sdk so they need to be able to pass this value in from their own `.env`. This just serves as an example of how to do that.

pauljung14 (approved):

PR #395, opened 2024-05-22
[MOB-8639]: add elements to OOTB type

mprew97:
## JIRA Ticket(s) if any

* [MOB-8639](https://iterable.atlassian.net/browse/MOB-8639)

## Description

Adds `elements` object to OOTB type to consolidate params. 

## Test Steps

[MOB-8639]: https://iterable.atlassian.net/browse/MOB-8639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/banner/index.tsx
+++ b/src/components/banner/index.tsx
@@ -21,7 +21,7 @@ const emptyElement = {
 export function IterableEmbeddedBanner({
   appPackageName,
   message,
-  elements = {
+  htmlElements = {
     parent: emptyElement,
     img: emptyElement,
     title: emptyElement,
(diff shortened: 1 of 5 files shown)


src/components/banner/index.tsx
```diff
@@ -21,15 +21,17 @@ const emptyElement = {
...
-  parent = emptyElement,
-  img = emptyElement,
-  title = emptyElement,
-  primaryButton = emptyElement,
-  secondaryButton = emptyElement,
-  body = emptyElement,
-  buttonsDiv = emptyElement,
-  textTitle = emptyElement,
-  textTitleImg = emptyElement,
+  elements = {
```
pauljung14, line 24: Just a thought, but hopefully this `elements` property doesn't get muddled/confused down the line with `message.elements` since both appear near each other, I wonder if maybe there's another name we could give this "elements" that something to do with the fact that we're pulling styles from them 🤔

pauljung14 (approved):
One concern about naming but otherwise lgtm

src/components/banner/index.tsx
mprew97: hmmmm good callout... will think on this some before merging.

PR #410, opened 2024-06-19
Fixed single item matches code

hardikmashru:
## JIRA Ticket(s) if any

## Test Steps


mprew97 (changes requested):
lgtm! can we add unit test cases to validate that the issue this is fixing is resolved and so it catches breaking changes in the future?

mprew97 (approved):

PR #416, opened 2024-07-05
Isset purchase update fix

hardikmashru:
## JIRA Ticket(s) if any

* [MOB-8943](https://iterable.atlassian.net/browse/MOB-8943)

## Description
- Added support for isSet
- Added unit tests for isSet

### DIFF
--- a/src/anonymousUserTracking/criteriaCompletionChecker.ts
+++ b/src/anonymousUserTracking/criteriaCompletionChecker.ts
@@ -334,7 +334,7 @@ class CriteriaCompletionChecker {
         if (eventName === UPDATE_CART && field === eventName) {
           return true;
         }
-        if (eventName !== UPDATE_CART && field === eventName) {
+        if (field === eventName) {
           return true;
         }
       }
--- a/src/anonymousUserTracking/criteriaCompletionChecker.ts
+++ b/src/anonymousUserTracking/criteriaCompletionChecker.ts
@@ -325,6 +325,19 @@ class CriteriaCompletionChecker {
     }
     const matchResult = filteredSearchQueries.every((query: any) => {
       const field = query.field;
+      if (
+        query.dataType === TRACK_EVENT &&
+        query.fieldType === 'object' &&
+        query.comparatorType === 'IsSet'
+      ) {
+        const eventName = eventData[KEY_EVENT_NAME];
+        if (eventName === UPDATE_CART && field === eventName) {
+          return true;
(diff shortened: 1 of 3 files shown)


src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -323,9 +344,19 @@ class CriteriaCompletionChecker {
   }
 
   private doesItemMatchQueries(item: any, searchQueries: any[]): boolean {
-    const filteredSearchQueries = searchQueries.filter((searchQuery) =>
-      Object.keys(item).includes(searchQuery.field)
-    );
+    const filteredSearchQueries = [];
+    for (let i = 0; i < searchQueries.length; i++) {
```
mprew97, line 348: I know we need to add some lint rules to prevent some of these back and forths we have on PRs, but please use array methods over for loops when possible.

src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -439,6 +470,16 @@ class CriteriaCompletionChecker {
     }
   }
 
+  private issetCheck(matchObj: any): boolean {
```
mprew97, line 473: If this is a json obj, 

```suggestion
  private issetCheck(matchObj: Record<any,any>): boolean {
```

mprew97 (changes requested):
Few minor change requests, otherwise looks good. Thanks for the tests!!

src/anonymousUserTracking/criteriaCompletionChecker.ts
harrymash2006: Done

src/anonymousUserTracking/criteriaCompletionChecker.ts
harrymash2006: not a json object.

mprew97: then what is it? would prefer to be explicit here

hardikmashru:
matchObj can be anything (string, boolean, double, integer)

src/anonymousUserTracking/criteriaCompletionChecker.ts
mprew97: please properly type this. after that is done, can merge.

mprew97 (approved):

PR #421, opened 2024-07-23
[MOB-9066]: better linting rules

mprew97:
## JIRA Ticket(s) if any

* [MOB-9066](https://iterable.atlassian.net/browse/MOB-9066)

## Description

Adds new linting rules
- extends `airbnb-base` https://www.npmjs.com/package/eslint-config-airbnb-base
- fixes a lot of linting issues
- also disables a lot to drastically reduce the blast radius
- this should at least make additional code cleaner and we will need to come back and clean up places where we are disabling the linter.

## Test Steps

[MOB-9066]: https://iterable.atlassian.net/browse/MOB-9066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/inapp/request.ts
```diff
@@ -37,100 +30,9 @@ type RequestMessagesProps = {
...
 };
 
-export const requestMessages = async ({ payload }: RequestMessagesProps) => {
-  /** @note TBD: Caching implementation and associated parameter will be enabled once new endpoint is ready */
+export const requestMessages = async ({ payload }: RequestMessagesProps) =>
+  /** @note TBD: Caching implementation and associated parameter
+   * will be enabled once new endpoint is ready */
   // if (!options?.useLocalCache) return await requestInAppMessages({});
   /** @note Always early return until then */
-  return await requestInAppMessages({ payload });
```
pauljung14, line 44: Everything after the early return here is valid and tested code, ready to be used when we have a new endpoint to handle cached messages. Think the linter doesn't like having unreachable code, so maybe we can just comment this out with a note? Or refactor to pull this out to a util?

pauljung14 (commented):
Pls see comment re: `src/inapp/request.ts` in particular

pauljung14 (approved):

PR #427, opened 2024-08-07
[MOB-9258] fixed nested IsSet matching

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9258](https://iterable.atlassian.net/browse/MOB-9258)

## Description
[Web SDK] nested object isSet does not match

## Test Steps

[MOB-9258]: https://iterable.atlassian.net/browse/MOB-9258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -357,18 +357,26 @@ class CriteriaCompletionChecker {
 
   private getValueFromNestedObject(eventData: any, field: string): any {
     const valueFromObj = this.getFieldValue(eventData, field);
-    if (typeof valueFromObj === 'object') {
-      return Object.keys(valueFromObj).map((key) =>
-        this.getValueFromNestedObject(valueFromObj, key)
-      );
+    if (typeof valueFromObj === 'object' && valueFromObj !== null) {
```
mprew97, line 360: Can we add additional test coverage for this method?

src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -357,18 +357,26 @@ class CriteriaCompletionChecker {
...
+      }, undefined);
     } else {
       return valueFromObj;
     }
   }
 
   private getFieldValue(data: any, field: string): any {
     const fields = field.split('.');
-    return fields.reduce((acc, field) => acc?.[field], data);
+    return fields.reduce((value, currentField) => {
```
mprew97, line 375: Same here as above

mprew97 (changes requested):

src/anonymousUserTracking/criteriaCompletionChecker.ts
darshan-iterable: While reviewing the scenario for writing the test case, I realised that this scenario will never occur because our portal doesn't accept values as objects when creating criteria. Therefore, I have removed that method, and everything is working fine for all the criteria.

mprew97 (approved):

PR #439, opened 2024-08-22
MOB-9308: supports nested field types

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9308](https://iterable.atlassian.net/browse/MOB-9308)

## Description
Support nested field types

## Test Steps

[MOB-9308]: https://iterable.atlassian.net/browse/MOB-9308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/anonymousUserTracking/criteriaCompletionChecker.ts
+++ b/src/anonymousUserTracking/criteriaCompletionChecker.ts
@@ -346,6 +346,25 @@ class CriteriaCompletionChecker {
       }
 
       if (field.includes('.')) {
+        let fields = field.split('.');
+        if (Array.isArray(eventData[fields[0]])) {
+          return eventData[fields[0]]?.every((item: any) => {
+            const data = {
+              [fields[0]]: item,
+              eventType: query?.eventType
+            };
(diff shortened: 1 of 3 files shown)


src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -346,6 +346,25 @@ class CriteriaCompletionChecker {
       }
 
       if (field.includes('.')) {
+        const fields = field.split('.');
```
mprew97, line 349: Why are we only getting the first `field` value? Shouldn't we run this compute on all `fields`?

darshan-iterable: as we have to take data from the array that's why we are taking only first value.

for example we have field `furniture.furnitureColor` and we have `furniture` as an array from the frontend,
so we have to check that `furniture` is an array or not, if it is an array then we are checking the each element of an array that matches or not

mprew97 (approved):

PR #440, opened 2024-08-23
fully supports comparison for data in Array data with all comparator

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9146](https://iterable.atlassian.net/browse/MOB-9146)

## Description
Fully supports comparison for data in Array data with all comparator types

[MOB-9146]: https://iterable.atlassian.net/browse/MOB-9146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/anonymousUserTracking/tests/compareArrayDataTypes.test.ts
+++ b/src/anonymousUserTracking/tests/compareArrayDataTypes.test.ts
@@ -1,0 +1,1448 @@
+import { SHARED_PREFS_EVENT_LIST_KEY } from '../../constants';
+import CriteriaCompletionChecker from '../criteriaCompletionChecker';
+
+const localStorageMock = {
+  getItem: jest.fn(),
+  setItem: jest.fn(),
+  removeItem: jest.fn()
+};
+
+describe('compareArrayDataTypes', () => {
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


src/anonymousUserTracking/tests/compareArrayDataTypes.test.ts
```diff
@@ -0,0 +1,1448 @@
...
+  beforeEach(() => {
+    (global as any).localStorage = localStorageMock;
+  });
+
+  // MARK: Equal
+  it('should return criteriaId 285 (compare array Equal)', () => {
+    (localStorage.getItem as jest.Mock).mockImplementation((key) => {
+      if (key === SHARED_PREFS_EVENT_LIST_KEY) {
+        return JSON.stringify([
+          {
```
mprew97, line 20: For these test cases that are using the search queries, we should create a file that contains constants of all of these mocked search queries. That way, the tests are easier to parse/grok and the search queries can be reused if needed.

mprew97 (changes requested):

PR #445, opened 2024-09-02
updated nested field logic

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9308](https://iterable.atlassian.net/browse/MOB-9308)

## Description
Updated logic same as Android [PR comments](https://github.com/Iterable/iterable-android-sdk/pull/778#discussion_r1737260636) to maintain consistency in functionality

[MOB-9308]: https://iterable.atlassian.net/browse/MOB-9308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #446, opened 2024-09-04
MOB-9145: support isOneOf and isNotOneOf comparator

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9145](https://iterable.atlassian.net/browse/MOB-9145)

## Description
Support isOneOf and isNotOneOf comparator

## Test Steps

[MOB-9145]: https://iterable.atlassian.net/browse/MOB-9145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


darshan-iterable:
This CI failed due to nested field, once this PR https://github.com/Iterable/iterable-web-sdk/pull/445 merged, CI will be success

src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -362,7 +362,7 @@ class CriteriaCompletionChecker {
               return this.evaluateComparison(
                 query.comparatorType,
                 valueFromObj,
-                query.value ? query.value : ''
+                query.value ? query.values : ''
```
mprew97, line 365: why is this one slightly different than the others?

src/anonymousUserTracking/criteriaCompletionChecker.ts
```diff
@@ -374,14 +374,14 @@ class CriteriaCompletionChecker {
           return this.evaluateComparison(
             query.comparatorType,
             valueFromObj,
-            query.value ? query.value : ''
+            query.value ?? query.values ?? ''
```
mprew97, line 377: what happens if an empty string is evaluated? should we even run the evaluation at all if `value` or `values` isn't present?

src/anonymousUserTracking/criteriaCompletionChecker.ts
darshan-iterable: Because in value we have single element and in values we have multiple element(s),
and when we create criteria for the isOneOf and isNotOneOf comparator we get values instead of value

src/anonymousUserTracking/criteriaCompletionChecker.ts
darshan-iterable: So empty string is compared with the criteria value, empty string is not matched with value and gives result as false. 
 Yes, we run the evaluation.

mprew97 (approved):

mprew97 (approved):

PR #450, opened 2024-09-09
[MOB-9505] rename merge parameter

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9505](https://iterable.atlassian.net/browse/MOB-9505)

## Description
Renamed merge parameter to `disableMergeAndEventReplay`

[MOB-9505]: https://iterable.atlassian.net/browse/MOB-9505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/authorization/authorization.ts
```diff
@@ -878,10 +920,16 @@ export function initialize(
         };
       };
       try {
-        merge = getMergeDefaultValue(merge);
-        const result = await tryMergeUser(userId, false, merge);
+        disableMergeAndEventReplay = getMergeDefaultValue(
```
davidtruong, line 923: Can we remove this logic for merging the user in setEmail and setUserId? We won't merge the user within the setUserID function. We will move the merge specific logic its own function.

The param should be disableEventReplay which will only affect whether the event gets replayed when going from an anonymous user to a known user.

darshan-iterable: yes removed and updated param

mprew97 (approved):

PR #453, opened 2024-09-11
[MOB-9402] update user should not be a separate call

darshan-iterable:
## JIRA Ticket(s) if any

* [MOB-9402](https://iterable.atlassian.net/browse/MOB-9402)

## Description
update user should not be a separate call

## Test Steps

[MOB-9402]: https://iterable.atlassian.net/browse/MOB-9402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mprew97 (approved):

mprew97 (changes requested):
Can we add a test case?

PR #463, opened 2024-10-14
[MOB-9899]: Add Auth Checks Before API Calls

mprew97:
## JIRA Ticket(s) if any

* [MOB-9899](https://iterable.atlassian.net/browse/MOB-9899)

## Description

Adds `typeOfAuth` verification before API calls. With AUT, it's possible folks try to make these calls before being initialized more frequently. Should be helpful regardless of AUT. 

## Test Steps

[MOB-9899]: https://iterable.atlassian.net/browse/MOB-9899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/embedded/embeddedManager.ts
```diff
@@ -27,8 +28,10 @@ export class IterableEmbeddedManager {
     callback: () => void,
     placementIds?: number[]
   ) {
-    await this.retrieveEmbeddedMessages(packageName, placementIds || []);
-    callback();
+    if (typeOfAuth !== null) {
+      await this.retrieveEmbeddedMessages(packageName, placementIds || []);
+      callback();
+    }
```
ts-nguyen, line 34: should we throw an error in the case of no auth?

ts-nguyen (approved):
I forget that js modules are basically singletons so attributes are global like this 😅 but looks good, just left one callout!

PR #475, opened 2024-10-31
[MOB-10091]: AUT Pre-Bash Updates

mprew97:
## JIRA Ticket(s) if any

* [MOB-10091](https://iterable.atlassian.net/browse/MOB-10091)

## Description
Renames config var, cleans up sample app without JWT fixes

## Test Steps

[MOB-10091]: https://iterable.atlassian.net/browse/MOB-10091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react-example/src/components/LoginFormWithoutJWT.tsx
+++ b/react-example/src/components/LoginFormWithoutJWT.tsx
@@ -2,6 +2,7 @@
 import { ChangeEvent, FC, FormEvent, useState } from 'react';
 import styled from 'styled-components';
 
+import { IdentityResolution } from '@iterable/web-sdk';
 import { TextField as _TextField } from './TextField';
 import { Button as _Button } from './Button';
 
@@ -39,7 +40,10 @@ const Error = styled.div`
 
 interface Props {
   setEmail: (email: string) => Promise<void>;
-  setUserId: (userId: string, merge?: boolean) => Promise<void>;
+  setUserId: (
+    userId: string,
+    identityResolution?: IdentityResolution
+  ) => Promise<void>;
   logout: () => void;
 }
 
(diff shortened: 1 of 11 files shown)


react-example/src/components/LoginFormWithoutJWT.tsx
```diff
@@ -62,7 +66,7 @@ export const LoginFormWithoutJWT: FC<Props> = ({
 
     const setUser = useEmail ? setEmail : setUserId;
 
-    setUser(user, true)
+    setUser(user)
```
ts-nguyen, line 69: I'm still unsure how identityResolution is being used, should we expect it to be passed in?

ts-nguyen (approved):
LGTM

react-example/src/components/LoginFormWithoutJWT.tsx
mprew97: Here, identity resolution _could_ be passed in. The config defaults are set to `mergeOnAnonymousToKnown: true` so just removing the variable here because it's redundant. The values for `IdentityResolution`, i.e., `mergeOnAnonymousToKnown` and `replayOnVisitorToKnown` could be passed in and changed any time `setEmail` or `setUserId` is called.

ts-nguyen: Ah so it's an override and there's logic in the implementation that pulls from the config? Could we label it as something like identityResolutionOverride in the function signatures where it would be an override?

ts-nguyen: Ah so it's an override and there's logic in the implementation that pulls from the config? If so, could we label it as something like identityResolutionOverride in the function signatures where it would be an override?

PR #558, opened 2026-04-07
Remove lodash and uuid dependencies to reduce bundle size

### LINKED ISSUE #442: Reduce bundle size of the SDK and/or make it treeshakable
The bundle size of the Iterable SDK minified is [294kb](https://bundlejs.com/?q=%40iterable%2Fweb-sdk%401.1.1) which is very large, some apps are smaller than this.

One way to reduce size of the library would be to make it more treeshable, if for example some feature is not used it could be stripped from the final bundle, maybe they could be opt in, for example if users only use tracking and not notifications and so on.

davidtruong:
Replace lodash/set with a simple object spread (single usage in inapp.ts) and replace uuid with native crypto.randomUUID() plus a fallback, removing ~73KB of unnecessary dependency weight. Fixes #442.

https://claude.ai/code/session_01EkgzU53JWSx3YHoupSiixd

## JIRA Ticket(s) if any

## Test Steps

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,10 +39,8 @@
     "buffer": "^6.0.3",
     "copy-webpack-plugin": "^11.0.0",
     "idb-keyval": "^6.2.0",
-    "lodash": "^4.17.21",
     "qs": "^6.12.3",
     "throttle-debounce": "^3.0.1",
-    "uuid": "^9.0.0",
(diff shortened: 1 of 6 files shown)


davidtruong:
Things are fixed

PR #559, opened 2026-04-07
Fix generateJWT closure capture to use updatable global reference

### LINKED ISSUE #160: `generateJWT` method does not respect function changes
# Description
The endpoint we use for fetching a JWT for iterable is itself behind a jwt authorizer - when this user token expires it is unknown how to pass this change to the `generateJWT` function already initialized in iterable. 

Code example:
```js
import { initialize } from '@iterable/iterable-web-sdk'

const IterableProvider = () => {
    const enabled = useFeatureFlag('iterable')
    const { userToken } = useUserToken();
    
    const handler = useCallback(async (request) => fetchJWT(token, request), [userToken])
    
    const iterableFunctions = useMemo(() => {
        if (enabled) {
            initialize(apiKey, handler)
        }
    }, [enabled])
}
```

in this scenario when the userToken expires and a new iterable JWT needs to be requested the jwt request will fail because it continues to use the expired token. Other workarounds I've tried are using a ref for the function, calling initialize again when the update happens, but none of these will update the generateJWT method.

Possible solution would just be giving us access to some kind of `setGenerateJWT` functionality.

davidtruong:
The generateJWT callback passed to initialize() was captured in a closure and never updated, causing stale references in React apps where the callback depends on changing state (e.g., rotating user tokens).

Changed all internal references from the closure-captured parameter to the module-level generateJWTGlobal variable, and added a setGenerateJWT() method to the WithJWT interface so users can update the function after initialization without reinitializing the entire SDK.

Also improved the type of generateJWTGlobal from `any` to a proper type alias (GenerateJWTFunction) for better type safety, and added jest.clearAllTimers() in test beforeEach to fix timer leak between tests.

Fixes #160

https://claude.ai/code/session_01TDMxsnQDkRoXn4181Vrv95

## JIRA Ticket(s) if any

## Test Steps

### DIFF
--- a/src/authorization/authorization.test.ts
+++ b/src/authorization/authorization.test.ts
@@ -71,6 +71,7 @@ describe('API Key Interceptors', () => {
   });
 
   beforeEach(() => {
+    jest.clearAllTimers();
     setTypeOfAuthForTestingOnly('userID');
 
     mockRequest.onPost('/users/update').reply(200, {
@@ -1190,6 +1191,7 @@ describe('User Identification', () => {
 
     describe('refreshJwtToken', () => {
       beforeEach(() => {
+        jest.clearAllTimers();
         mockRequest.resetHistory();
       });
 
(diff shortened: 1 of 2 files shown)


davidtruong:
Things are fixed

PR #561, opened 2026-04-07
Add allowIframeScripts config to fix Safari in-app message functionality

### LINKED ISSUE #490: inApp clicks Safari
### Problem

inApp (inBrowser) messages interaction still works differently in Safari. No javascript results in no click tracking and no option for custom actions `action://...`. This restriction means it is basically restricted on all browsers when using a single campaign with custom actions.

Why is this still the case? Adding `allow-scripts` to the iframe sandbox attr, and removing the specific safari checks in the inApp code makes it work perfectly. What's the reason for not adding `allow-scripts` then? The content is controlled by iterable + us so where are the security concerns? Any other browser runs it anyway?

davidtruong:
Safari blocks JS execution in sandboxed iframes without allow-scripts, preventing click tracking, action:// links, and proper close button behavior. This adds an opt-in allowIframeScripts config option that adds allow-scripts to the iframe sandbox attribute and disables Safari-specific workarounds, giving Safari users the same full functionality as Chrome/Firefox.

Fixes #490

https://claude.ai/code/session_01Lnv6rRRdYi5TVqPDpjVgvT

## JIRA Ticket(s) if any

## Test Steps

### DIFF
--- a/README.md
+++ b/README.md
@@ -1947,6 +1947,7 @@ type Options = {
   enableUnknownActivation: boolean;
   isEuIterableService: boolean;
   dangerouslyAllowJsPopups: boolean;
+  allowIframeScripts: boolean;
   eventThresholdLimit?: number;
   onUnknownUserCreated?: (userId: string) => void;
   identityResolution?: {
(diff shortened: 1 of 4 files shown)


davidtruong:
Things are fixed

PR #563, opened 2026-04-07
fix: move dev-only deps to devDependencies and make customPayload opt…

### LINKED ISSUE #528: copy-webpack-plugin should be a dev dependency
It is only used during development thus not needed for users to install as a dependency.
https://github.com/Iterable/iterable-web-sdk/blob/3948ff5f53bee310f5005d0b0b94bdeac5ee991b/webpack.dev.config.js#L1-L5

davidtruong:
…ional

Move copy-webpack-plugin and axios-mock-adapter from dependencies to devDependencies since they are only used in webpack dev config and test files respectively. This prevents consumers from installing unnecessary packages, including the vulnerable serialize-javascript transitive dep.

Also make InAppMessage.customPayload optional to match runtime behavior where the API may return undefined when no custom payload is configured.

Fixes #528, Fixes #538

https://claude.ai/code/session_011XYfVrCRu4H8y24ZEpoboX

## JIRA Ticket(s) if any

## Test Steps

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,7 @@
     "@pabra/sortby": "^1.0.1",
     "@types/ws": "8.5.4",
     "axios": "1.14.0",
-    "axios-mock-adapter": "^1.22.0",
     "buffer": "^6.0.3",
-    "copy-webpack-plugin": "^11.0.0",
     "idb-keyval": "^6.2.0",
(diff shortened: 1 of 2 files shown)


davidtruong:
Things are fixed

PR #567, opened 2026-04-07
Fix trackInApp* payload mutation and add JWT auth test coverage (#552)

davidtruong:
Replace delete statements with destructuring to avoid mutating the caller's payload object. The interceptor already overwrites email/userId with the correct values from SDK state, so the delete was both unnecessary and harmful as a side effect. Also adds missing JWT auth test coverage for in-app event endpoints.

https://claude.ai/code/session_016eGrVtXnTGForEz4J3Krrm

## JIRA Ticket(s) if any

## Test Steps

### DIFF
--- a/src/authorization/authorization.test.ts
+++ b/src/authorization/authorization.test.ts
@@ -934,6 +934,57 @@ describe('User Identification', () => {
         );
       });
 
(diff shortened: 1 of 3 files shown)


davidtruong:
Things are fixed

PR #576, opened 2026-04-28
SDK-412 UUA Naming inconsistencies

joaodordio:
## JIRA Ticket(s) if any

* [SDK-412](https://iterable.atlassian.net/browse/SDK-412)

## Description
Aligns the Web SDK with the cross-SDK Unknown User Activation (UUA) naming convention. No behavior changes. Old public surface kept as deprecated aliases; legacy on-disk data is migrated transparently on first read.

## Renames

- Config flag: `enableUnknownActivation` is now `enableUnknownUserActivation`. Old name kept as a deprecated alias; if both are supplied, the new one wins. Setting the legacy name emits a `console.warn`.
- localStorage key: `unknown_userId` is now `itbl_userid_unknown` (matches the `itbl_*` convention used by every other Web key).
- Session storage JSON fields inside `itbl_unknown_sessions`: `number_of_sessions` / `first_session` / `last_session` are now `totalUnknownSessionCount` / `firstUnknownSession` / `lastUnknownSession` (already what the payload types and backend expect; the rename layer in `createUnknownUser` is gone).
- Constant: `AUA_WARNING` is now `UUA_WARNING`. Old export kept as a deprecated alias.

## Migrations

- New `migrateLegacyKey(oldKey, newKey)` helper in `utils/commonFunctions.ts`, invoked once at SDK init for the unknown user id key.
- `updateUnknownSession` reads legacy snake_case fields as fallback and rewrites under the new camelCase keys.

## Tests

- New `commonFunctions.test.ts` covering the migration helper (idempotent, no overwrite of new key, no-op when neither exists).
- New tests in `config.test.ts` for the deprecated config alias path.
- New session field migration test in `unknownUserEventManager.test.ts`.
- README and `react-example` updated.

[SDK-412]: https://iterable.atlassian.net/browse/SDK-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/authorization/authorization.ts
+++ b/src/authorization/authorization.ts
@@ -7,6 +7,7 @@ import {
   IS_PRODUCTION,
   STATIC_HEADERS,
   SHARED_PREF_UNKNOWN_USER_ID,
+  LEGACY_SHARED_PREF_UNKNOWN_USER_ID,
   RouteConfig,
   SHARED_PREFS_CRITERIA,
   SHARED_PREF_CONSENT_TIMESTAMP,
@@ -14,6 +15,7 @@ import {
   SHARED_PREF_USER_ID,
   RETRY_USER_ATTEMPTS
 } from '../constants';
+import { migrateLegacyKey } from '../utils/commonFunctions';
 import { UnknownUserMerge } from '../unknownUserTracking/unknownUserMerge';
 import {
   UnknownUserEventManager,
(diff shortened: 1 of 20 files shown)


src/authorization/authorization.ts
```diff
@@ -396,6 +398,11 @@ export function initialize(
 ) {
   apiKey = authToken;
   generateJWTGlobal = generateJWT;
+  // One-shot migration from the legacy non-prefixed unknown user id key.
+  migrateLegacyKey(
```
mprew97, line 402: Do we not need to do the same with the fields renamed in `itbl_unknown_sessions`?

mprew97 (commented):
Mostly looks good here, but i'm slightly concerned about in process sessions. Have we tested a running live session before/after these changes?

